PR #43, opened 2016-03-19
First Iteration of the debian packaging

martinnaughton:


ralight:
Before you go any further down this route, I suspect that the approach you're taking will be rejected by debian. In my experience the best approach is to use the "debhelper" method. I've picked a random debian python package that you can take a look at: https://anonscm.debian.org/cgit/debian-science/packages/pyfftw.git/tree/debian

martinnaughton:
ok cool. i will change the way i am doing it. Sorry for the few pull request. It thinks i am doing a pull request as my other user.

PR #47, opened 2016-03-20
First go at packaging source code in to a debian package. Committing …

martinnaughton:
…all files in case they are needed later. After the review and everything is full working i will remove any unused files. It successfully builds a package with dpkg-buildpackage

```
This is the package layout. It is the same as the previous python-mosquitto package
/usr/share/doc/paho.mqtt.python/changelog.Debian.gz
/usr/share/doc/paho.mqtt.python/copyright
/usr/share/doc/paho.mqtt.python/README.Debian
/usr/share/doc/paho.mqtt.python/changelog.gz
/usr/share/pyshared/client.py
/usr/share/pyshared/publish.py
/usr/share/pyshared/paho_mqtt-1.1.egg-info
```


martinnaughton:
finally got it up and ready to review.

I will remove unused files once everything is reviewed.

A few issues i had
- dh_usrlocal was failing because it expects empty directories if it finds usr/local in the directory path. 
- the make file for the project was putting all the files from the build directory in to the package directory. I did not want the __init__.py files in the package so i override the build and make it install in in directory usr in the project root. Then used install file to copy the files i wanted to the package directory. Not sure if there is a better way to do this then what i have done.

PierreF:
I'm not sure we should use the Makefile when building package. It could avoid issue with dh_usrlocal or the need to manual install.
In addition, the file are wrongly placed with current install. client.py is installed in /usr/share/pyshared/client.py, which will ends up being imported with "import client", not "import paho.mqtt.client"

This could be fixed by specifying the buildsystem to dh. In rules:
```                                        
%:                                                                              
    dh $@ --with python2 --buildsystem=pybuild
```

To works, control file will also need to add a Build-Depends on python-all (anyway it's a good idea :))


Also once using pybuild, it should be rather easy to have package for python2 and python3.

Finally, since the Python package is named "paho-mqtt" I think that packages should be "python-paho-mqtt" and "python3-paho-mqtt" to match all other Python package. In such case source package is usually "paho-mqtt". I don't known if that name is the good one, since "paho-mqtt" could be used by other language.

PierreF:
Closing as this PR seems staled. Feel free to reopen if you are still working on it.

PR #49, opened 2016-04-07
Support configuring size of in/out message queue

changyuheng:


src/paho/mqtt/client.py
```diff
@@ -897,7 +899,13 @@ def publish(self, topic, payload=None, qos=0, retain=False):
             message.dup = False
 
             self._out_message_mutex.acquire()
+
+            if self._max_out_messages > 0 and len(self._out_messages) >= self._max_out_messages:
+                self._out_message_mutex.release()
+                return (MQTT_ERR_NOMEM, local_mid)
```
ralight: MQTT_ERR_NOMEM isn't an appropriate error code here.

src/paho/mqtt/client.py
```diff
@@ -897,7 +899,13 @@ def publish(self, topic, payload=None, qos=0, retain=False):
...
+
             self._out_messages.append(message)
+            print('size of self._out_messages:', len(self._out_messages))
```
ralight: I think this has been left in by mistake.

src/paho/mqtt/client.py
```diff
@@ -2124,6 +2148,8 @@ def _handle_publish(self):
         elif message.qos == 2:
             rc = self._send_pubrec(message.mid)
             message.state = mqtt_ms_wait_for_pubrel
+            if self._max_in_messages > 0 and len(self._in_messages) >= self._max_in_messages:
```
ralight: I think returning an error here will disconnect the client, which seems like a bad outcome.

ralight:
I've put some specific comments in the code, but have more general comments as well.

First off, all new feature changes should be made on the develop branch not master.
Secondly, you need to sign the Eclipse CLA to contribute to this repository and add a signed-off line to your commit message - see CONTRIBUTING.md.

I think adding this for the outgoing queue isn't a bad idea. For consistency with other projects I'd suggest using "max_queued_messages" for naming. It needs a new error code like I suggested in the line comment.

For incoming messages your current solution is to disconnect the client if the broker sends "too many" messages. This is a bad idea. There's not a lot else we can do though, we've got to keep processing the incoming data. With the current MQTT protocol I don't believe it is useful to attempt to control incoming data flow. I won't accept that part at the moment.

changyuheng:
I on traced the outgoing queue and guess incoming queue has the same mechanism. Sorry for not verified. The print line is for debugging, you're right. I'll modify this let it dedicate for outgoing message queue with following contribution guide. Thank you.

src/paho/mqtt/client.py
changyuheng: What's the appropriate error code. Could you please give me a name? I'm not an English speaker.

changyuheng:
I follow the instruction and created another pull request.
https://github.com/eclipse/paho.mqtt.python/pull/50

So I'll close this one.

PR #51, opened 2016-04-11
Clean up stray exceptions following broker disconnect with TLS

swails:
Python 3.4 started raising ValueError instead of AttributeError when writing to
a disconnected SSL socket. When the client writes some stray messages following
a call to disconnect(), they are not caught in Python 3.4 or Python 3.5.

Python Issue: https://bugs.python.org/issue9177

Python Commit: https://mail.python.org/pipermail/python-checkins/2013-July/123012.html

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -1325,10 +1325,10 @@ class Client(object):
         Messages that match 'sub' will be passed to 'callback'. Any
         non-matching messages will be passed to the default on_message
         callback.
-        
+
         Call multiple times with different 'sub' to define multiple topic
         specific callbacks.
-        
+
         Topic specific callbacks may be removed with
         message_callback_remove()."""
         if callback is None or sub is None:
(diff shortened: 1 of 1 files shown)


ralight:
Thanks, this looks good to me.

Could you please just amend the commit message to add a "Signed-off-by" line to the end of the commit as would be produced with "git commit -s".

swails:
[It already has that in the commit log](https://github.com/eclipse/paho.mqtt.python/pull/51/commits/1b8efcc6f03a84be9fb90480c69a97346ff35b3b).

Do you also want it in the PR text?

Thanks!

swails:
I also want to bring your attention to the change of permissions in `client.py` to remove execute permissions.  I didn't see any `if __name__ == '__main__':` or anything else that suggested this file would ever be directly executed (even if it did have a `if __name__ == '__main__'`, the canonical way to execute a module within a package is via `python -m`).

But it's easy to miss that subtle change, so I wanted to point it out explicitly.

ralight:
Sorry, my mistake, I wasn't looking properly!

I agree with the executable flag as well - I was surprised to notice it myself a little while ago.

Issue #53, opened 2016-04-20
Callbacks definition and inheritance


AlexisBRENON:
Hi guys,

I'm using the Python implementation of MQTT through the paho-mqtt package. The documentation says that you can inherit from the Client class to create your own MQTT_client. Nevertheless, inheriting from the client doesn't allow you to define some on_* functions directly in the class because they are overridden by the __init__ call of the base class...

So, there is any particular reason the define Client.on_* to None in the __init__ instead of using classical member functions like this :

```python
class Client(object):
    def __init__(self, ...):
        ...
        # Nothing about self.on_*

    def on_*(self, ...):
        pass
```

An implementation like this would not break the actual behavior (you can redefine the functions, using mqttc.on_connect = my_new_connect_function), and it would be easier to inherit, defining overriding on_* functions in the sub-class.

If this seems a good idea, I propose myself to do the patch.

Kind regards,
Alexis BRENON

AlexisBRENON:
Roger Light answers:

> Hi Alexis,
> 
> That sounds like a good idea. I confess that inheritance isn't
> something I really use so it may have been possible in the past and
> I've broken it.
> 
> I'd be happy to accept a patch for this, please base it on the fixes
> branch, add an entry to the changelog and make sure you have signed
> the Eclipse CLA.
> 
> Thanks,
> 
> Roger

AlexisBRENON:
Jason Swails answers:

> Hi,
> 
> To be fair, the current class structure doesn't prevent inheritance -- just like the base class, you need to explicitly attach any callbacks as an attribute to the class.  Anything that's not the on_<action> callback is implemented directly as a method attribute, and can be overridden just fine in a subclass.
> 
> But this is definitely a strange pattern in my experience with Python.  Most instance attributes are assigned when the class definition is defined and is an unbound instance attribute on the class (rather than something that is bound at instance initialization).  A more natural API (to me, of course!) would be to either leave the on_* callbacks not implemented and rely on subclassing to implement it -or- have a function where you can register the callbacks.  The former would be a substantial break in the API.  The latter could be implemented in a fully backwards-compatible way by making the on_<action> callbacks descriptors (via the @property decorator)... but that's really the same thing that we already have from a functional perspective... it would just allow overriding to work in subclasses.
> 
[...]

PR #54, opened 2016-04-20
Callbacks definition and inheritance 

AlexisBRENON:
Related to issue #53.

### DIFF
--- a/examples/sub-class.py
+++ b/examples/sub-class.py
@@ -4,9 +4,9 @@
 #
 # All rights reserved. This program and the accompanying materials
 # are made available under the terms of the Eclipse Distribution License v1.0
-# which accompanies this distribution. 
+# which accompanies this distribution.
 #
-# The Eclipse Distribution License is available at 
+# The Eclipse Distribution License is available at
 #   http://www.eclipse.org/org/documents/edl-v10.php.
 #
 # Contributors:
@@ -29,36 +29,30 @@ except ImportError:
         sys.path.insert(0, cmd_subfolder)
     import paho.mqtt.client as mqtt
 
-class MyMQTTClass:
-    def __init__(self, clientid=None):
-        self._mqttc = mqtt.Client(clientid)
-        self._mqttc.on_message = self.mqtt_on_message
-        self._mqttc.on_connect = self.mqtt_on_connect
-        self._mqttc.on_publish = self.mqtt_on_publish
-        self._mqttc.on_subscribe = self.mqtt_on_subscribe
+class MyMQTTClass(mqtt.Client):
 
-    def mqtt_on_connect(self, mqttc, obj, flags, rc):
+    def on_connect(self, mqttc, obj, flags, rc):
         print("rc: "+str(rc))
 
-    def mqtt_on_message(self, mqttc, obj, msg):
+    def on_message(self, mqttc, obj, msg):
         print(msg.topic+" "+str(msg.qos)+" "+str(msg.payload))
 
-    def mqtt_on_publish(self, mqttc, obj, mid):
+    def on_publish(self, mqttc, obj, mid):
         print("mid: "+str(mid))
 
-    def mqtt_on_subscribe(self, mqttc, obj, mid, granted_qos):
+    def on_subscribe(self, mqttc, obj, mid, granted_qos):
         print("Subscribed: "+str(mid)+" "+str(granted_qos))
 
-    def mqtt_on_log(self, mqttc, obj, level, string):
+    def on_log(self, mqttc, obj, level, string):
         print(string)
 
     def run(self):
-        self._mqttc.connect("m2m.eclipse.org", 1883, 60)
-        self._mqttc.subscribe("$SYS/#", 0)
+        self.connect("m2m.eclipse.org", 1883, 60)
+        self.subscribe("$SYS/#", 0)
 
         rc = 0
         while rc == 0:
-            rc = self._mqttc.loop()
+            rc = self.loop()
         return rc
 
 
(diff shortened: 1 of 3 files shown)


swails:
I'm -0.5 on this.  I won't call it a full -1, but I don't think there is a strong enough justification for the break in the API, personally.  What you're doing is replacing `None` by a no-op function call.  These are *very* different beasts, and this can break some common debug code out in the wild right now.  Consider the use-case of having some kind of debug-activating function that attaches callbacks to any MQTTClient that doesn't have one already:

```Python
def set_debug(list_of_clients):
    for client in list_of_clients:
        if client.on_publish is None: client.on_publish = debug_on_publish
        if client.on_connect is None: client.on_connect = debug_on_connect
...
```

You would do this if you wanted to avoid replacing any existing callbacks with your debug callback (for whatever reason, maybe for some kind of polling, or some very cursory debugging of a specific part of the code).  With this change, there is no easy way to tell whether a callback has been implemented or not.

To properly implement the inheritance you hope to gain from this PR, all you'd have to do is bind `self.on_<method>` to an actual function inside your subclass's initializer.

So I guess my question here is: has the API been declared stable yet?  The 1.1 version tag suggests it has (standard versioning uses 0.x for unstable APIs).  If so, this PR violates that tacit guarantee in ways I can't fully predict.

I think there is a fairly simple way of allowing the inheritance behavior you seek *without* breaking backwards compatibility.  IMO, that should be explored first.

swails:
See AlexisBRENON/paho.mqtt.python#1.  Together with that commit, we get the full behavior implemented here in addition to (I believe) full backwards-compatibility.  Even if it's decided to cut out the "old" way of specifying callbacks (to go back to "one, and preferably only one obvious way to do something"), this will at least permit adding a deprecation warning with a transition period where users are warned of the impending break, rather than an abrupt breakage.

During the process of working on that PR, I stumbled upon some initial Python behavior I thought was rather surprising at first.  That is, if you attach a function that is not an instance method to one of the callbacks, you don't have access to `self` unless it is explicitly passed as a parameter (which it happens to be in the case of these callbacks, but the call signature of an instance method needs an extra `self` argument while the call signature of any other function must omit `self`).  I think @AlexisBRENON's approach here has somewhat fewer surprises, so I'm switching my vote to +0.8 (pushed up to a full +1 if we retain backwards compatibility).

AlexisBRENON:
Merged work from @swails . This preserves API compatibility, allowing classic inheritance pattern too.

examples/sub-class.py
```diff
@@ -29,36 +29,30 @@
...
-    def __init__(self, clientid=None):
-        self._mqttc = mqtt.Client(clientid)
-        self._mqttc.on_message = self.mqtt_on_message
-        self._mqttc.on_connect = self.mqtt_on_connect
-        self._mqttc.on_publish = self.mqtt_on_publish
-        self._mqttc.on_subscribe = self.mqtt_on_subscribe
+class MyMQTTClass(mqtt.Client):
 
-    def mqtt_on_connect(self, mqttc, obj, flags, rc):
+    def on_connect(self, mqttc, obj, flags, rc):
```
Nzbuu: Should the `mqttc` argument be removed from here and the other callbacks?

swails: No, that would break backwards-compatibility.  Look where `on_connect` is called -- `self` is explicitly passed as the first argument, so `mqttc` is still required.

Nzbuu: Right, I see. You end up with `on_connect(self, self, ...)` effectively. That's pretty ugly. Is there any way to make it so that the inheritance-based interface doesn't need the repeated argument.

Might it be better to create an intermediate method, say `self.handle_connect(*args)`, as a regular method, that has a default implementation that calls `self.on_connect(self, *args)`.

swails: > Is there a way to make it so that the inheritance-based interface doesn't need the repeated argument?

The only way to do this in a way that works in a backwards-compatible way (i.e., so you can tack on a callback via simple attribute assignment) is by using introspection techniques to query the number of arguments that the function takes and making your decision based on that.

But this breaks if you pass a decorated function that is generalized to `*args` and `**kwargs` (as many decorators are).  So this is a very fragile approach.

I agree that `on_connect(self, self, ...)` is ugly, but the solutions to work around that *and* maintain backwards compatibility are worse.  If it's the second `self` that's annoying you, you can always make it a `staticmethod`.  You get rid of the superfluous `self` (since all you use is `mqttc` anyway), but you still have to explicitly pass `self` when called in an instance.  But then anyone making subclasses would have to remember that those callbacks either need an additional `self` or a `@staticmethod` decorator.

ralight: @Nzbuu If you can come up with something that is better/less complex than passing a single extra argument that is still backwards compatible with both inherited and non-inherited cases then I'd be all for it.

Nzbuu: So, I've been thinking a lot about this, and, I now think that the current implementation (with or without this pull request) is the "right" behaviour, since, in this scenario where an instance method is used as the callback, the first argument is the "application" object and the second is the "MQTT client" object.

In particular, IMO, implementing the callbacks by inheriting from the `Client` class is a "bad" design choice and a "better" design is to use composition, i.e. your application should "use" a MQTT client object rather than "be" a MQTT client, since this gives better encapsulation and SoC. You could even use DI to inject the MQTT client. This is how the "subclassing" example currently works, except that `MyMqttClass` should be `MyMqttApp` or similar, and I would suggest to maintain that example as a "composition" example.

I'm sorry to interfere; I'm just an interested member of the community.

ralight: To clear up a point of confusion, to give "sub-class.py" it's full name you'd write "an-example-of-a-subscribing-client-that-uses-a-class.py", not "an-example-of-a-client-using-subclassing.py". I think I'll rename it to be a bit clearer.

It's also clear that using an MQTT client object rather than inheriting is the most common pattern in use.

ralight:
Thanks for this work, I've now manually merged it into the develop branch so am going to close this request.

PR #62, opened 2016-05-30
Update client.py

yoch:
Dont overwrite callbacks if exists.

Allows to attach callbacks simply by subclassing, like :

```python
class MyClient(mqtt.Client):
    def on_message(client, userdata, msg):
        pass
        pass
```


ralight:
Have you taken a look at the work done on the develop branch in this area?

yoch:
No, I just see (why to use property here ?). But it doesn't really solve my problem.

Also, I realized that my solution is wrong (bound vs unbound method) unless we use `staticmethod`.

Currently, I use this hack, but it's a little bit ugly :

```python
class MyClient(mqtt.Client):
    def __init__(self, *args, **kwargs):
        # init client
        super().__init__(*args, **kwargs)
        # rebind callbacks
        for name, attr in self.__class__.__dict__.items():
            if name.startswith('on_'):
                self.__dict__[name] = attr
```

yoch:
Okay, I found the full story : https://github.com/eclipse/paho.mqtt.python/pull/54

IMHO, it's quite surprising to write code like : 

```python
class MyMQTTClass(mqtt.Client):
    def on_message(self, mqttc, obj, msg):
        print(msg.topic+" "+str(msg.qos)+" "+str(msg.payload))
```

(note that `self` and `mqttc` is the same object)

ralight:
Yes, agreed (and that was raised on the original conversation). The more common way of using the module is creating an instance of a client rather than inheriting. This function definition allows both modes of operation.

PR #63, opened 2016-05-31
Update client.py

yoch:
Bind all callbacks as instance methods.

To improve https://github.com/eclipse/paho.mqtt.python/pull/54 , allows to subclass `Client` properly without duplicating `self`.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -37,6 +37,7 @@ import time
 import uuid
 import base64
 import hashlib
+import types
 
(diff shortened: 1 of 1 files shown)


yoch:
A drawback with my current solution is in case an user subclass the client and defines callback, and then attach another callback with the usual method. I'm looking for a proper solution for that.

PierreF:
This issue seems stalled. Are you still working on it or should we close this PR ?

yoch:
You can close this PR.

PR #64, opened 2016-06-01
Replace _out_packet list by collections.deque

robertobarreda:
- deque provides higher perfomance popping from left end of the list

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,93 +1,79 @@
-Contributing to Paho
-====================
+# Contributing to Paho
 
-Thanks for your interest in this project.
+Thanks for your interest in this project!
 
-Project description:
---------------------
+You can contribute bugfixes and new features by sending pull requests through GitHub.
 
-The Paho project has been created to provide scalable open-source implementations of open and standard messaging protocols aimed at new, existing, and emerging applications for Machine-to-Machine (M2M) and Internet of Things (IoT).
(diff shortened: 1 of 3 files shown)


ralight:
I've got conflicting pull requests #64 and #69. I think either is better than the current code, but can we discuss which is best?

robertobarreda:
@ralight I think both #64 and #69 changes are compatible:
- #69 changes _out_messages and _in_messages
- #64 changes _out_packet

You can merge one of the pull requests and then we can fix the other one based on the last version of the code.

yoch:
@ralight, @robertobarreda yes, these PR are fully compatible with #69, because #64 apply to `_out_packet` and #69 to `_out_messages` and `_in_messages`.

karlp:
Same issue applies to publish.multiple, only there it still misses the pop(0) to publish in the order of the original list

Nzbuu:
I think the target branch for this pull request should be "develop" not "master"

ralight:
This looks fine, but the ip validation checks are not passing. Please sign the Eclipse ECA with the email that you use for your commits. There are some commits that have a different email address of robertobarreda@users.noreply.github.com which I assume it's actually yours... Please also add a signed-off-by line to your commit (use "git commit -s").

Finally, as you've got to redo the commits anyway could you try and make it without the extra merges and existing commits.

Thanks for your patience!

robertobarreda:
@ralight I've created a new pull-request #116 to properly merge with develop branch and fix the signed-off-by issue. I'm closing this issue.

PR #66, opened 2016-06-01
Use RLock instead of Lock for _callback_mutex (fix #12)

### LINKED ISSUE #12: Calling message_callback_add within a callback causes deadlock
migrated from Bugzilla [#459012](https://bugs.eclipse.org/bugs/show_bug.cgi?id=459012)
status UNCONFIRMED severity *normal* in component *MQTT-Python* for *1.1*
Reported in version *1.1* on platform *Other*
Assigned to: Roger Light

On 2015-02-03 06:09:31 -0500, Ed Morris wrote:

> The following code locks up when a message is published to the topic 'bar'.
> 
> #!/usr/bin/python
> 
> import paho.mqtt.client as paho
> 
> def on_foo_message(mqttc,request,msg):
>     print("foo on_message")
>     pass
> 
> def on_message(mqttc,request,msg):
>     print("global on_message")
>     mqttc.subscribe("foo", 0)
>     print("Subscribed to foo")
>     mqttc.message_callback_add("foo/+",on_foo_message)
>     print("Added callback to foo")
> 
> mqttc = paho.Client()
> broker = "127.0.0.1"
> mqttport = 1883
> mqttc.on_message = on_message
> mqttc.connect(broker, mqttport, 60)
> mqttc.subscribe("bar", 0)
> mqttc.loop_forever()
> 
> I think it is because the call to self._callback_mutex.acquire() in message_callback_add never exits. This is because the mutex was already acquired in _handle_on_message and will not release until the on_message callback has finished - which cannot happen.

On 2015-06-10 22:38:22 -0400, YuLun Shih wrote:

> Same issue here, I think it's all about _callback_mutex

yoch:
Fix for https://github.com/eclipse/paho.mqtt.python/issues/12

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -476,7 +476,7 @@ class Client(object):
         self._bind_address = ""
         self._in_callback = False
         self._strict_protocol = False
-        self._callback_mutex = threading.Lock()
+        self._callback_mutex = threading.RLock()
         self._state_mutex = threading.Lock()
         self._out_packet_mutex = threading.Lock()
         self._current_out_packet_mutex = threading.Lock()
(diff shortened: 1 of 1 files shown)


ralight:
Thanks for this, sorry I'd already fixed and tested it based on your comment on the bug but hadn't pushed the change.

Issue #67, opened 2016-06-02
topic_matches_sub() incorrectly filter some subscriptions


yoch:
`('sport/#', 'sport/')` -> `False` instead of `True`

Also, some invalid filters are accepted.

`('sport+', 'sport')` -> `True` which is incorrect. etc

Here my proposal, which is simpler than the previous code, can detect invalid subscriptions, and runs faster (by about 30%) :

``` python
def topic_matches_sub(sub, topic):
    "return None if the subscription is invalid (maybe raise an exception ?)"

    if not sub or not topic:
        # invalid (all topic names and topic filters MUST be at least one character long)
        return None

    if topic[0] == '$' and sub[0] != '$':
        return False

    sub_l, topic_l = sub.split('/'), topic.split('/')
    sub_sz, topic_sz = len(sub_l), len(topic_l)

    for i, level in enumerate(sub_l):

        if level == '#':
            if i + 1 < sub_sz:
                return None # invalid ('#' not at end)
            return True

        elif level == '+':
            if i >= topic_sz:
                return False

        else:
            if '#' in level or '+' in level:
                return None # invalid ('#' / '+' in level part)
            if i >= topic_sz or level != topic_l[i]:
                return False

    return sub_sz == topic_sz
```

On the other hand, it doesn't detect all inconsistencies in topic format.

ralight:
> On the other hand, it doesn't detect all inconsistencies in topic format.

Could you explain the inconsistencies you are referring to?

ralight:
Ah, you mean it detects invalid subscriptions but not invalid topics?

yoch:
I mean MQTT **topics** constraints :

- The wildcard characters [...] MUST NOT be used within a Topic Name [MQTT-4.7.1-1].
- Topic Names and Topic Filters MUST NOT include the null character (Unicode U+0000) [MQTT-4.7.3-2]

It would be nice to have dedicated functions to validate inputs, but it's not the job for `topic_matches_sub()`.

ralight:
Agreed on both counts. I'll merge this after 1.2 is released.

yoch:
Here a check for subscriptions filters :

```python
    def _filter_wildcard_len_check(sub):
        if (len(sub) == 0 or len(sub) > 65535
            or any('+' in p or '#' in p for p in sub.split('/') if len(p) > 1)
            or '#/' in sub):
            return MQTT_ERR_INVAL
        else:
            return MQTT_ERR_SUCCESS
[...]

PR #68, opened 2016-06-04
use a topic matcher for better performance

yoch:
I wrote a new class called `MQTTMatcher`for managing callbacks. Also fix https://github.com/eclipse/paho.mqtt.python/issues/67

Here a simple benchmark (1000 topics and 1000 corresponding subscriptions) : 

```python
from random import *
from string import ascii_lowercase as charset
from time import time

N = 1000

def gen(sz):
    return '/'.join(choice(charset) for _ in range(sz))

def to_filter(topic):
    lst = topic.split('/')
    i = randrange(0, len(lst))
    if random() < 0.5:
        lst[i] = '+'
    else:
        lst[i:] = ['#']
    return '/'.join(lst)


topics = [gen(randint(2,7)) for _ in range(N)]
filters = [to_filter(topic) for topic in topics]

matcher = MQTTMatcher()
for f in filters:
    matcher[f] = f

t1 = time()
dct1 = {t: set(matcher.subscribers(t)) for t in topics}
t2 = time()
print('MQTTMatcher: matching %d topics in %gs' % (N, t2 - t1))

t1 = time()
dct2 = {}
for t in topics:
    dct2[t] = set()
    for f in filters:
        if topic_matches_sub(f, t):
            dct2[t].add(f)

t2 = time()
print('topic_matches_sub: matching %d topics in %gs' % (N, t2 - t1))

assert dct1 == dct2
```

The result is very eloquent : 

```
MQTTMatcher: matching 1000 topics in 0s
topic_matches_sub: matching 1000 topics in 1.48453s
```

This is because current implementation is in O(N) for each topic, which are tested against all subscriptions.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -199,6 +199,23 @@ def connack_string(connack_code):
         return "Connection Refused: unknown reason."
 
 
+def topic_matches_sub(sub, topic):
+    """Check whether a topic matches a subscription.
+
+    For example:
+
+    foo/bar would match the subscription foo/# or +/bar
+    non/matching would not match the subscription non/+/+
+    """
+    matcher = MQTTMatcher()
+    matcher[sub] = True
+    try:
+        next(matcher.iter_match(topic))
+        return True
+    except StopIteration:
+        return False
+
+
 def _socketpair_compat():
     """TCP/IP socketpair including Windows support"""
     listensock = socket.socket(socket.AF_INET, socket.SOCK_STREAM, socket.IPPROTO_IP)
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -50,6 +50,8 @@ try:
 except ImportError:
     HAVE_DNS = False
 
+from .matcher import MQTTMatcher
+
 if platform.system() == 'Windows':
     EAGAIN = errno.WSAEWOULDBLOCK
 else:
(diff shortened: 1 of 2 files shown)


src/paho/mqtt/client.py
```diff
@@ -1630,11 +1560,10 @@ def message_callback_remove(self, sub):
             raise ValueError("sub must defined.")
 
         self._callback_mutex.acquire()
-        for i in range(0, len(self.on_message_filtered)):
-            if self.on_message_filtered[i][0] == sub:
-                self.on_message_filtered.pop(i)
-                self._callback_mutex.release()
-                return
+        try:
+            del self._on_message_filtered[sub]
```
robertobarreda: self._on_message_filtered.pop(sub, None)

yoch: `MQTTMatcher` doesn't have a `pop` method (this can be added, of course).

robertobarreda: sorry, my fault

src/paho/mqtt/client.py
```diff
@@ -1630,11 +1560,10 @@ def message_callback_remove(self, sub):
...
+        except:  # no such subscription
```
robertobarreda: except KeyError: (avoid capture all exceptions)

yoch: You're right.

ralight:
I've taken a look at this and it's good, but the function `topic_matches_sub()` has been removed so the tests fail. The function was also a public function so other code may rely on it. Could you please reimplement it using the matcher class?

yoch:
Okay. So I

- rebased with latest develop branch
- assembled all previous commits in one
- restored `topic_matches_sub` using `MQTTMatcher` (we also can use the function I wrote in #67)

ralight:
Thanks for your contribution!

PR #69, opened 2016-06-06
use OrderedDict for _out_messages and _in_messages

yoch:
I think this is more suitable, and should improve performance in some cases.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -37,6 +37,7 @@ import time
 import uuid
 import base64
 import hashlib
+import collections
 
 HAVE_DNS = True
 try:
@@ -515,8 +516,8 @@ class Client(object):
         self._ping_t = 0
         self._last_mid = 0
         self._state = mqtt_cs_new
-        self._out_messages = []
-        self._in_messages = []
+        self._out_messages = collections.OrderedDict()
+        self._in_messages = collections.OrderedDict()
         self._max_inflight_messages = 20
         self._inflight_messages = 0
         self._max_queued_messages = 0
@@ -990,9 +991,9 @@ class Client(object):
                 self._out_message_mutex.release()
                 return (MQTT_ERR_QUEUE_SIZE, local_mid)
 
-            self._out_messages.append(message)
+            self._out_messages[message.mid] = message
             if self._max_inflight_messages == 0 or self._inflight_messages < self._max_inflight_messages:
-                self._inflight_messages = self._inflight_messages+1
+                self._inflight_messages += 1
                 if qos == 1:
                     message.state = mqtt_ms_wait_for_puback
                 elif qos == 2:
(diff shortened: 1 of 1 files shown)


src/paho/mqtt/client.py
```diff
@@ -2137,13 +2138,13 @@ def _message_retry_check_actual(self, messages, mutex):
         mutex.release()
 
     def _message_retry_check(self):
-        self._message_retry_check_actual(self._out_messages, self._out_message_mutex)
-        self._message_retry_check_actual(self._in_messages, self._in_message_mutex)
+        self._message_retry_check_actual(self._out_messages.values(), self._out_message_mutex)
```
robertobarreda: accessing to self._out_messages and self._in_messages .values() must be done inside _message_retry_check_actual after locking the mutex

yoch: indeed, I didn't notice that it could be a problem with python2

yoch:
may slow down things on python 2, due to the use of `values()`

yoch:
A workaround is to use this conditional function instead of the `values()` method : 

```python
if sys.version_info < (3,):
    itervalues = lambda dct: dct.itervalues()
else:
    itervalues = lambda dct: dct.values()
```

Nzbuu:
@yoch , I think it's better to use the `six` module for this sort of backwards compatibility (https://pythonhosted.org/six/index.html#six.itervalues)

swails:
It's easy to bundle `six` - many projects do it :smiley:

PierreF:
Should we bundle six or use a real dependencies ?

I would opt for real dependencies, I find it cleaner and six is very broadly used, so it should be easy to install on target systems (python-six/python3-six is present on Debian 7, Ubuntu 12.04, CentOS 7, ...).

PierreF:
Also while speaking of performance, iterating over a OrderedDict is costly on Python2 (about 20 times more than a list)
Also itervalues() don't speed up thing notably.
```
>>> import timeit, sys
>>> sys.version
'2.7.13 (default, Jan 19 2017, 14:48:08) \n[GCC 6.3.0 20170118]'
>>> timeit.timeit('for m in messages: pass', 'import collections; messages=[]\nfor i in range(500000): messages.append(i)', number=100)
0.6489040851593018
>>> timeit.timeit('for m in messages.values(): pass', 'import collections; messages=collections.OrderedDict()\nfor i in range(500000): messages[i]=i', number=100)
11.83427095413208
>>> timeit.timeit('for m in messages.itervalues(): pass', 'import collections; messages=collections.OrderedDict()\nfor i in range(500000): messages[i]=i', number=100)
11.830179929733276
```

Due to such small difference between values() and itervalues() I would always use values().

But since the cost of iterating is 20 times more than a list, is it really worse moving to OrderedDict ?

ralight:
To add a dependency that isn't part of the Python core platform we must get the dependency checked. This is done through ipzilla: https://dev.eclipse.org/ipzilla/

It shouldn't be a big deal with a MIT license and we can decide whether to ask to distribute as well, or not.

From the point of view of Linux distributions, bundling would not be the preferred choice, they would want to maintain the other package separately.

PierreF:
Hum... while speaking of IP checks and dependencies, does test dependencies should also go though ipzilla ? (I think to pytest, pytest-runner, pylama, ...).

PR #70, opened 2016-06-06
Fix 01-con-discon-success.test

yoch:
Use version 3.1.1 by default


ralight:
This is already done, so I'm going to close this request.

PR #71, opened 2016-06-06
adding some logs for debug

yoch:
Adding seemingly missing logs


ralight:
Yes... I don't know why they aren't there.

Could you add information about what is being sent to the log lines - e.g. what the keepalive value is, whether username/password are set, that sort of thing.

Please also base it on the develop branch.

Issue #73, opened 2016-06-08
allow Client Identifier to be a nullstring


yoch:
Why using no client_id [isn't authorized](https://github.com/eclipse/paho.mqtt.python/blob/master/src/paho/mqtt/client.py#L499) ? It's sometime useful, and this is allowed by MQTT specifications : 

> A Server MAY allow a Client to supply a ClientId that has a length of zero bytes, however if it does so the Server MUST treat this as a special case and assign a unique ClientId to that Client. It MUST then process the CONNECT packet as if the Client had provided that unique ClientId [MQTT-3.1.3-6].
> 
> If the Client supplies a zero-byte ClientId, the Client MUST also set CleanSession to 1 [MQTT-3.1.3-7].

___

Also, current implementation of random ClientId is not very safe, because it can lead to subtle bugs. For instance, this script causes `mqtt.Client` chooses the same ClientId for clients, introducing competition between them to take over if we run several clients : 

```python
import random
import paho.mqtt.client as mqtt


def on_connect(client, userdata, flags, rc):
    print(client._client_id, "Connected with result code ", rc)

def on_disconnect(client, userdata, rc):
    print(client._client_id, "Disconnected with result code ", rc)


# suppose, for some reason, we make
random.seed(0)

# and just later start the client
client = mqtt.Client()
client.on_connect = on_connect
client.on_disconnect = on_disconnect

client.connect("iot.eclipse.org")
client.loop_forever()
```

ralight:
It's largely historical, MQTT 3.1 didn't allow zero length client ids.

ralight:
I'm treating this as a bug for zero length client ids, which is now fixed. Your pull request should then fix the randomness (please use separate bug reports for separate issues to make it easier to deal with).

Thanks for your work.

PR #74, opened 2016-06-08
fix for #73

yoch:
fix for #73

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -37,6 +37,7 @@ import threading
 import time
 import uuid
 import base64
+import string
 import hashlib
 try:
     # Use monotionic clock if available
@@ -195,6 +196,17 @@ def connack_string(connack_code):
         return "Connection Refused: not authorised."
     else:
         return "Connection Refused: unknown reason."
+
+
+def base62(num, base=string.digits+string.ascii_letters, padding=1):
+    """Convert a number to base-62 representation."""
+    assert num >= 0
+    digits = []
+    while num:
+        num, rest = divmod(num, 62)
+        digits.append(base[rest])
+    digits.extend(base[0] for _ in range(len(digits), padding))
+    return ''.join(reversed(digits))
 
 
 def topic_matches_sub(sub, topic):
@@ -499,9 +511,10 @@ class Client(object):
         self._message_retry = 20
         self._last_retry_check = 0
         self._clean_session = clean_session
+
         if client_id == "" or client_id is None:
             if protocol == MQTTv31:
-                self._client_id = "paho/" + "".join(random.choice("0123456789ADCDEF") for x in range(23-5))
+                self._client_id = base62(uuid.uuid4().int, padding=22)
             else:
                 self._client_id = ""
         else:


yoch:
I doubt if the default behavior should be to generate a random client_id or use a empty client_id. But current behavior (mixing the two concepts) seems incorrect.

ralight:
The point with generating an id rather than using None is that it works
with both 3.1 and 3.1.1.

I like the uuid approach, but there is no need to truncate to 23 characters
these days.

yoch:
I'm not familiar with the 3.1 version. But forcing users to use client_id random, even if he wants a null one, may be problematic. So this PR seems to be the right approach.

Reason to limit to 23 characters is to be as compatible as possible : 

> The Server **MUST** allow ClientIds which are between 1 and 23 UTF-8 encoded bytes in length, and that contain only the characters
> "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ" [MQTT-3.1.3-5].
>  
> The Server **MAY** allow ClientId’s that contain more than 23 encoded bytes. The Server MAY allow ClientId’s that contain characters not included in the list given above.

michaelplaing:
uuid's are 128 bits and can be compacted to fit in 23 text bytes.

I suspect that is where that limit came from.

michaelplaing:
i.e. using base 62.

(this is as I recall from a test some years ago)

michaelplaing:
we import 'numconv' to assist with base 62 conversions, which we use for
another purpose.

yoch:
@michaelplaing 
- 128 bits can be encoded in 22 alphanumeric characters in base 62 : *ceil((128 * log(2)) / log(62))*
- this can be added easyly if necessary (thus avoiding dependencies)

```python
import string 
import uuid

def base62(num, base=string.digits+string.ascii_letters, padding=1):
	assert num >= 0
	digits = []
	while num:
		num, rest = divmod(num, 62)
		digits.append(base[rest])
	digits.extend(base[0] for _ in range(len(digits), padding))
	return ''.join(reversed(digits))

base62(uuid.uuid4().int, padding=22)
```

ralight:
That looks good to me, feel free to add it to the current request.

I'm told the 23 character limit is down to a field in WebSphere MQ being limited to 24 characters - so 23 characters plus a zero terminator.

michaelplaing:
Thanks for the history, Roger!

ralight:
If you update this pull request to just include the uuid changes I'll accept it.

If clean session is set to False and the user does not provide a client id, they have no way of knowing what the client id is (ok, so they could look inside their client instance for the random id, but it's hardly good behaviour) and so they are creating a durable client that can never be resumed. This is not good.

yoch:
I agree. But it's still possible to allow null client id with clean session, right ?

ralight:
"With clean session" means "clean_session=True" to me. In that case you can do what you want. If clean_session is False, the client id cannot be zero length going to the broker, and the library intentionally prevents you from doing that if has generated a random id for you.

ralight:
Sorry I wasn't clear, could you scrub the rest of the history (i.e. a force push without the other commits) - this only needs to be a single commit. Let's keep things tidy :)

yoch:
done

PR #76, opened 2016-06-10
logs improvements

yoch:
same as #71 on develop branch

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -2081,9 +2081,11 @@ class Client(object):
                 self._pack_str16(packet, self._password)
 
         self._keepalive = keepalive
+        self._easy_log(MQTT_LOG_DEBUG, "Sending CONNECT")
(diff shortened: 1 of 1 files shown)


yoch:
I added the possibility to use _eay_log with formatters, which makes the code much readable.

Now, we can write code like : 

```python
self._easy_log(MQTT_LOG_DEBUG, "Sending PUBLISH (d%d, q%d, r%d, m%d), '%s', ... (%d bytes)", 
  dup, qos, retain, mid, topic, payloadlen)
```

instead of 
```python
self._easy_log(MQTT_LOG_DEBUG, "Sending PUBLISH (d"+str(int(dup))
  +", q"+str(qos)+", r"+str(int(retain))+", m"+str(mid)+"), '"
  +topic+"', ... ("+str(payloadlen)+" bytes)")
```

ralight:
Thanks, this is a good change.

PR #77, opened 2016-06-13
Code cleanup

yoch:
- many code simplifications
- python 2 / 3 integration made simpler
- improve some inputs handling

Also, some bugs were fixed, I plan to write about this later.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -2027,7 +2027,7 @@ class Client(object):
 
             connect_flags |= 0x04 | ((self._will_qos&0x03) << 3) | ((self._will_retain&0x01) << 5)
 
-        if self._username:
+        if self._username is not None:
             remaining_length += 2+len(self._username)
             connect_flags |= 0x80
             if self._password is not None:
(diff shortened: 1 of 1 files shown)


yoch:
fix #78

src/paho/mqtt/client.py
```diff
@@ -2009,25 +2014,25 @@ def _send_connect(self, keepalive, clean_session):
...
             protocol = PROTOCOL_NAMEv311
             proto_ver = 4
+
         remaining_length = 2+len(protocol) + 1+1+2 + 2+len(self._client_id)
         connect_flags = 0
         if clean_session:
             connect_flags |= 0x02
 
         if self._will:
+            remaining_length += 2+len(self._will_topic)
```
ralight: This substitution is wrong.

ralight:
This is good stuff, with a couple of minor exceptions.

Sorry to be awkward though, but could you split this out into separate commits by category of fix? I'm a simple man and like to keep things straightforward and easy to follow where possible. Something like "Replace all instances of x=x+y with x+=y"

"Replace all instances of this style of code:

x = topic
if x[0] == 0:
    ....

with

x, = topic
if x == 0:
    ...."

It's much easier to review that way.

I noticed that you've changed the behaviour a bit as well - you no longer do complete type checks on the payload data. I've not tested it, will that now automatically raise a ValueError?

ralight:
Closing this due to PR #81.

PR #81, opened 2016-06-16
Code cleanup (replace PR #77)

yoch:
This is a replacement for #77, split by categories of fix as @ralight asked.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -331,7 +331,7 @@ class MQTTMessageInfo:
     def wait_for_publish(self):
         """Block until the message associated with this object is published."""
         with self._condition:
-            while self._published == False:
+            while not self._published:
                 self._condition.wait()
 
     def is_published(self):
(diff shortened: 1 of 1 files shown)


yoch:
NOTE: the last commit will not work on python3 without PR #76, but I will kept them separated for now.

src/paho/mqtt/client.py
```diff
@@ -958,27 +966,29 @@ def publish(self, topic, payload=None, qos=0, retain=False):
...
+
+        if qos < 0 or qos > 2:
             raise ValueError('Invalid QoS level.')
-        if isinstance(payload, str) or isinstance(payload, bytearray):
-            local_payload = payload
-        elif sys.version_info[0] == 3 and isinstance(payload, bytes):
-            local_payload = bytearray(payload)
-        elif sys.version_info[0] < 3 and isinstance(payload, unicode):
+
+        if isinstance(payload, unicode):
```
Nzbuu: Feels like there's lots of duplication of the encoding functionality. Would it be better to have a single encode method and re-use it for every conversion from input type to bytes? Especially for the message payloads and the will payloads. But also for topic names, user names, passwords, etc.

yoch: not all conversions have the same needs (and semantic), and I have already removed many duplicates. but this may be a good idea for message and will payloads.

Nzbuu: I haven't read the specification in detail yet (I will this week) but certainly there should be common code for encoding and decoding the publish payloads and the length-prefixed strings.

PR #83, opened 2016-06-21
check subscriptions filters

### LINKED ISSUE #67: topic_matches_sub() incorrectly filter some subscriptions

yoch:
fix #67

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -1075,6 +1075,9 @@ class Client(object):
         if topic_qos_list is None:
             raise ValueError("No topic specified, or incorrect topic type.")
 
+        if any(self._filter_wildcard_len_check(topic) != MQTT_ERR_SUCCESS for topic, _ in topic_qos_list):
+            raise ValueError('Invalid subscription filter.')
+
         if self._sock is None and self._ssl is None:
             return (MQTT_ERR_NO_CONN, None)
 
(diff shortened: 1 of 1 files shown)


ralight:
This has now been merged outside of the github pull request interface (due to the conflict), so I'm closing this request. Thanks for your contribution!

PR #84, opened 2016-06-21
simplifying ssl

yoch:
I don't understand the need to manage two sockets for the ssl support (this makes the code more complicated).

Here an attempt to make it simpler with assigning `self._sock` to the ssl wrapper directly. Tested and apparently works well.

If there is a mistake with this approach, please can you explain the reason to manage two separate sockets ?


ralight:
I think this is an artifact of porting from the original C code, where it wasn't possible to do this.

PR #86, opened 2016-06-23
use __slots__ for MQTTMessageInfo and MQTTMessage

yoch:
This can slightly reduce the memory usage.


ralight:
Thanks for this.

In the future could I please request that you add a comment to ChangeLog.txt as well? It's easier for others to keep track of changes there.

PR #87, opened 2016-06-24
helpers: publish messages in order

karlp:
This has "less" performance, but is likely to be a lot closer to what
people actually expect to happen, rather than the current behaviour of
publishing in reverse order.

Signed-off-by: Karl Palsson karlp@tweak.net.au

See also: https://github.com/eclipse/paho.mqtt.python/pull/64 which deal with the performance implications of lists vs deques for ordered popping.

### DIFF
--- a/src/paho/mqtt/publish.py
+++ b/src/paho/mqtt/publish.py
@@ -25,7 +25,7 @@ import paho.mqtt as mqtt
 
 def _do_publish(c):
     """Internal function"""
-    m = c._userdata.pop()
+    m = c._userdata.pop(0)
     if type(m) is dict:
         topic = m['topic']
         try:


karlp:
:facepunch: eclipse CLA  validation is bogus.  My commiter, author and s-o-b are all the same, and match my eclipse account.  It claims that "karlp" (my github name) isn't matching.

PierreF:
To avoid the possible high cost of pop(0), wouldn't be simpler to reverse() the list before starting any processing ?
Once the list is reversed, we could do a normal pop() that is fast.

Nzbuu:
I think I agree with @PierreF . It would be better if the `_userdata` list were reversed first, then you can pop from the right.

Alternatively, it could use a `deque`, but I think that's unnecessary for this function since it's never adding extra items to the queue.

swails:
+1 for using `deque`, -1 for doing `list.reverse()`.  Many of my use-cases send a single message at a time.  In that case, `pop(0)` is the same as `pop()`, and avoiding the unnecessary call to `reverse()` makes the original a tiny bit faster (although admittedly not much).

For small lists (I've tested 1-3 elements), there is little difference between reverse/pop, pop(0), and deque.popleft().  For larger lists, pop(0) is by far the slowest and deque.popleft() is by far the fastest.

Basically, what I've found is that if the performance is worth reversing the list, then it's definitely worth using a deque instead of list altogether.  And it's not like a deque is really any harder or more inconvenient to use than a list...

Below is my timing comparison (the computer I ran it on is admittedly budget) in Python 3.5:

```python
In [1]: from collections import deque

In [2]: def f(length):
   ...:     l = list(range(length))
   ...:     while l:
   ...:         l.pop(0)
   ...: 
   ...: def f2(length):
   ...:     l = list(range(length))
   ...:     l.reverse()
   ...:     while l:
   ...:         l.pop()
   ...: 
   ...: def f3(length):
   ...:     l = deque(list(range(length)))
   ...:     while l:
   ...:         l.popleft()
   ...:         

In [3]: %timeit f(1)
   ...: %timeit f2(1)
   ...: %timeit f3(1)
The slowest run took 4.35 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 1.59 µs per loop
1000000 loops, best of 3: 2.79 µs per loop
100000 loops, best of 3: 2.19 µs per loop

In [4]: %timeit f(2)
   ...: %timeit f2(2)
   ...: %timeit f3(2)
   ...: 
100000 loops, best of 3: 3.93 µs per loop
100000 loops, best of 3: 2.39 µs per loop
100000 loops, best of 3: 3.03 µs per loop

In [5]: %timeit f(3)
   ...: %timeit f2(3)
   ...: %timeit f3(3)
   ...: 
100000 loops, best of 3: 4.06 µs per loop
100000 loops, best of 3: 3.67 µs per loop
100000 loops, best of 3: 3.74 µs per loop

In [6]: %timeit f(100)
   ...: %timeit f2(100)
   ...: %timeit f3(100)
   ...: 
10000 loops, best of 3: 51.9 µs per loop
10000 loops, best of 3: 32.1 µs per loop
10000 loops, best of 3: 22.7 µs per loop

```

Nzbuu:
@swails , I roughly get the same performance results. In fact, for any length, I get the `deque` version being the fastest in Python 3.5. So I agree that that using `deque` does seem better than using `reverse`. Although, I don't know if older versions will necessarily agree.

ralight:
I think targetting performance for the latest version of Python is appropriate.

PierreF:
Fixed by #215

PR #88, opened 2016-07-05
allow to set custom URL path for MQTT over WS usage

rac01:
The recent MQTT spec allows URL paths other than /mqtt when using a WebSocket transport. The suggested modification keeps the current behavior but allows
to set a custom URL path using
    endpoint_url_path_set(self, path):

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -531,6 +531,7 @@ class Client(object):
         self._will_retain = False
         self.on_message_filtered = []
         self._host = ""
+        self._path= "/mqtt"
         self._port = 1883
         self._bind_address = ""
         self._in_callback = False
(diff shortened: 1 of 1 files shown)


src/paho/mqtt/client.py
```diff
@@ -2644,12 +2648,13 @@ class WebsocketWrapper:
...
     OPCODE_PONG = 0xa
 
-    def __init__(self, socket, host, port, is_ssl):
+    def __init__(self, socket, host, path, port, is_ssl):
 
         self.connected = False
 
         self._ssl = is_ssl
         self._host = host
+        self._path = path
```
Nzbuu: Would it be better to store the encode path directly?

Nzbuu:
I think the target branch for this pull request should be "develop" not "master"

rac01:
I'll follow up on the "pull request should be for the develop branch" hint and re-submit there.

PR #90, opened 2016-07-08
Avoid SSL handshake to hang forever

### LINKED ISSUE #3: Reconnect on an SSL socket can block forever
migrated from Bugzilla [#478959](https://bugs.eclipse.org/bugs/show_bug.cgi?id=478959)
status UNCONFIRMED severity _major_ in component _MQTT-Python_ for _1.2_
Reported in version _unspecified_ on platform _PC_
Assigned to: Roger Light

Original attachment names and IDs:
- _[paho-client-TLS-reconnect-timeout.patch](https://bugs.eclipse.org/bugs//attachment.cgi?id=257015)_ (ID 257015)

On 2015-10-03 13:25:52 -0400, Jeff Jasper wrote:

> Created attachment 257015
> Patch for the bug.
> 
> The ssl.wrap_socket call within the reconnect method can block forever if the broker (for whatever reason) does not respond during the handshake. I don't know what's causing the broker (RabbitMQ) to not respond but it occurs occasionally and happens more frequently under network conditions with high latencies.
> 
> I'm including a patch where I perform the TLS handshake after wrapping the socket and give the socket a 10 second timeout. After the handshake completes I set the socket to non-blocking. Feel free to change my hard coded 10 sec. block to something more configurable.
> 
> Also included in the patch is a broad exception catch to release the _current_out_packet_mutex to avoid potential deadlock (I've never experienced this but noticed it when I was troubleshooting the previous issue).

On 2016-01-04 19:04:49 -0500, Adam mills wrote:

> We are seeing this as well with Mosquitto as a broker.

PierreF:
This PR should fix #3. It's the patch of Jeff Jasper slightly modified:
- Use keepalive as timeout instead of hard-coded 10 seconds
- Don't call setblocking(0), it's already done few line after
- Don't include the other fix about possible deadlock

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -829,7 +829,10 @@ class Client(object):
                 ca_certs=self._tls_ca_certs,
                 cert_reqs=self._tls_cert_reqs,
                 ssl_version=self._tls_version,
-                ciphers=self._tls_ciphers)
+                ciphers=self._tls_ciphers,
+                do_handshake_on_connect=False)
+            self._ssl.settimeout(self._keepalive)
+            self._ssl.do_handshake()
 
             if self._tls_insecure is False:
                 if sys.version_info[0] < 3 or (sys.version_info[0] == 3 and sys.version_info[1] < 2):


Nzbuu:
I think this conflicts with the changes in https://github.com/eclipse/paho.mqtt.python/pull/84

PierreF:
But #84 don't seems to fix the SSL handshake that could hang, so this PR is still needed

Nzbuu:
Agreed. I just wanted to highlight a merge conflict that is brewing that someone will have to fix so that @ralight can decide who he wants to fix it before he's forced to do it himself.

ralight:
Yes... there are a few alternative implementations in the pull requests. Makes life interesting :)

PierreF:
#124 fix this issue, but the PR is done by the original author.

Issue #75, opened 2016-06-08
client sometimes crash because the topic can't be decoded in utf8


yoch:
Sometimes I got this exception (and the client hangs) : 

```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "C:\Users\yoch\AppData\Local\Programs\Python\Python35-32\lib\threading.py", line 914, in _bootstrap_inner
    self.run()
  File "C:\Users\yoch\AppData\Local\Programs\Python\Python35-32\lib\threading.py", line 862, in run
    self._target(*self._args, **self._kwargs)
  File "C:/Users/yoch/Desktop/paho.mqtt/src\paho\mqtt\client.py", line 2567, in _thread_main
    self.loop_forever()
  File "C:/Users/yoch/Desktop/paho.mqtt/src\paho\mqtt\client.py", line 1367, in loop_forever
    rc = self.loop(timeout, max_packets)
  File "C:/Users/yoch/Desktop/paho.mqtt/src\paho\mqtt\client.py", line 891, in loop
    rc = self.loop_read(max_packets)
  File "C:/Users/yoch/Desktop/paho.mqtt/src\paho\mqtt\client.py", line 1169, in loop_read
    rc = self._packet_read()
  File "C:/Users/yoch/Desktop/paho.mqtt/src\paho\mqtt\client.py", line 1754, in _packet_read
    rc = self._packet_handle()
  File "C:/Users/yoch/Desktop/paho.mqtt/src\paho\mqtt\client.py", line 2227, in _packet_handle
    return self._handle_publish()
  File "C:/Users/yoch/Desktop/paho.mqtt/src\paho\mqtt\client.py", line 2385, in _handle_publish
    message.topic = message.topic.decode('utf-8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf4 in position 20: invalid continuation byte
```

To reproduce this bug, you can simply connect a client to `"iot.eclipse.org"` and subscribe to `"#"`, usually after a short time the crash occurs.

ralight:
Thanks, I've pushed a fix for this - the exception will now be raised when the user accesses message.topic in the on_message callback.

yoch:
According with the specification, it should rather close the connection (the broker is buggy). But I don't know what is the best choice in practice.

> If a Server or Client receives a Control Packet containing ill-formed UTF-8 it MUST close the Network Connection [MQTT-1.5.3-1].

andrewbernard:
[...]

PR #91, opened 2016-07-08
Fix publish on  Python3 with QoS > 0

PierreF:
#75 introduced a regression with Python 3 clients. When trying to publish a message with QoS > 0, the following error occurred:

```
  File "simple.py", line 4, in <module>
    client.publish('mytopic', 'mymessage', 1)
  File "/home/pierref/dev/ext/paho.mqtt.python/src/paho/mqtt/client.py", line 1022, in publish
    rc = self._send_publish(message.mid, message.topic, message.payload, message.qos, message.retain, message.dup)
  File "/home/pierref/dev/ext/paho.mqtt.python/src/paho/mqtt/client.py", line 370, in topic
    return self._topic.decode('utf-8')
AttributeError: 'str' object has no attribute 'decode'
```

This PR fix the issue by encoding topic to UTF-8 before creating the MQTTMessage object.

To reproduce the issue, i used this script (simple.py):

```

client = mqtt.Client()
client.publish('mytopic', 'mymessage', 1)
```

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -347,11 +347,13 @@ class MQTTMessage:
 
     Members:
 
-    topic : String. topic that the message was published on.
+    topic : String/bytes. topic that the message was published on.
     payload : String/bytes the message payload.
     qos : Integer. The message Quality of Service 0, 1 or 2.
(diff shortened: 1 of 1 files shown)


ralight:
Thanks for this, sorry for the delay in looking at it.

Issue #11, opened 2016-02-04
Paho does not support the SNI extension of TLS


jpwsutton:
migrated from Bugzilla [#471037](https://bugs.eclipse.org/bugs/show_bug.cgi?id=471037)
status UNCONFIRMED severity _enhancement_ in component _MQTT-Python_ for _1.2_
Reported in version _1.2_ on platform _All_
Assigned to: Roger Light

On 2015-06-25 14:04:34 -0400, Juan Carlos Jimenez wrote:

> The current implementation of Paho Python does not support the Server Name Indication extension of TLS.
> 
> This means that if a server has 2 VPS's trying to use TLS in the same port the connection will fail.
> 
> SNI support was added to the ssl module in the python version 2.7.9.
> 
> I have submitted a patch that will try to use SNI by default (pretty much every single SSL library out there supports it nowadays). If the ssl module does not support SNI, it will fall back to the non-SNI implementation.
> 
> You can find it here: https://git.eclipse.org/r/#/c/50658/
> 
> I'm currently using that patch and it is working fine for me.
> 
> Just opening the bug report for reference.

On 2015-06-25 17:50:34 -0400, Roger Light wrote:

> Thanks for the report and the gerrit push, I had spotted it I've just been quite busy recently. I'll try to have a look soon.

Lokesh-K-Haralakatta:
Why can't we merge the changes from this commit - https://github.com/exosite/paho.mqtt.python/commit/9acc14285b2c06b14816d017912c25ec9ff1f0da to master repo? This commit has the changes to support to SNI for paho.mqtt.python. I used the code changes for one of our connectivity to IoT Platform for Client Side Certificate support, the changes worked fine. Can we have this changes merged ASAP?

PierreF:
SNI support is already added on develop branch by PR #92 but not yet released.

Lokesh-K-Haralakatta:
Any idea, when can we have the SNI support released and available as part of paython pkg?

PierreF:
Closing the issue, since fix is already committed.
Release with SNI support is expected by end of June

PR #92, opened 2016-07-08
Add support for SSL context and SNI

Nzbuu:
Use SSLContext objects, if available, since this provides more options and features including SNI (#11). Also, allows default system context from `ssl.create_default_context()` to be used, improves encapsulation and supports dependency injection. 

**Update**: Removed support for SSL without SSLContext (#115)

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -24,10 +24,7 @@ import socket
 try:
     import ssl
 except ImportError:
-    HAVE_SSL = False
     ssl = None
-else:
-    HAVE_SSL = True
 
 import struct
 import sys
@@ -258,16 +255,6 @@ def _socketpair_compat():
     return (sock1, sock2)
 
 
-def _check_can_read_file(filename):
-    if filename:
-        try:
-            f = open(filename, "r")
-        except IOError as err:
-            raise IOError(filename + ": " + err.strerror)
-        else:
-            f.close()
-
-
 class MQTTMessageInfo:
     """This is a class returned from Client.publish() and can be used to find
     out the mid of the message that was published, and to determine whether the
@@ -446,6 +433,7 @@ class Client(object):
       MQTT_LOG_ERR, and MQTT_LOG_DEBUG. The message itself is in buf.
 
     """
+
     def __init__(self, client_id="", clean_session=True, userdata=None, protocol=MQTTv311, transport="tcp"):
         """client_id is the unique client id string used when connecting to the
         broker. If client_id is zero length or None, then the behaviour is
@@ -546,7 +534,7 @@ class Client(object):
         self._thread_terminate = False
         self._ssl = False
         self._ssl_context = None
-        self._tls_insecure = False
+        self._tls_insecure = False  # Only used when SSL context does not have check_hostname attribute
         self._logger = None
         # No default callbacks
         self._on_log = None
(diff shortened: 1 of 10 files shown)


Nzbuu:
I've rebased this on the latest `develop` branch to get the updated tests, etc.

anthonyrisinger:
How can we move this forward? Is more work or review needed? SNI is the greatest!

PierreF:
Just spotted a small unexpected behavious: tls_insecure_set() does not work when using tls_set_context() with a ssl.create_default_context().

This is caused by [SSLContext.check_hostname](https://docs.python.org/2.7/library/ssl.html#ssl.SSLContext.check_hostname) that already performs the [match_hostname](https://docs.python.org/2.7/library/ssl.html#ssl.SSLSocket.do_handshake) during handshake. This flag is due on a SSLContext produced by create_default_context().

Note: when using tls_set() and tls_insecure_set() it works since in this case the SSLContext is created using ssl.SSLContext() and not create_default_context().

I think either tls_insecure_set() should update the check_hostname flag on the ssl context or add a note in documentation that tls_insecure_set should not be used with tls_set_context.


Step to reproduce:

```
# Certificate created with: openssl req -x509 -nodes -newkey rsa:1024 -keyout server.key -out server.pem -days 365 -subj '/CN=localhost' -batch
# mosquitto config:
# listener 1885
# cafile server.pem
# certfile server.pem
# keyfile server.key

import paho.mqtt.client
import ssl

cl2 = paho.mqtt.client.Client()
ssl_ctx = ssl.create_default_context(cafile='server.pem')
ssl_ctx.check_hostname = False  # Without this line, I does not work
cl2.tls_set_context(ssl_ctx)
cl2.tls_insecure_set(True)

# certificate was generated for CN=localhost, not "CN=127.0.0.1"
cl2.connect('127.0.0.1', 1885)
```

Nzbuu:
@PierreF , that makes sense. I think it's just a matter of re-arranging some of the logic. I'll look at it in more detail later.

Nzbuu:
OK. So I think it's pretty complicated to get all of the logic completely consistent. I'm going open an issue to propose removing SSL support in Python versions that don't include SSLContext (i.e. require Python 2.7.9+ or Python 3.2+). It would remove quite a lot of branches and make the intended behaviour clearer.

ralight:
That sounds fine.

Nzbuu:
@PierreF , your issue should be fixed now

src/paho/mqtt/client.py
```diff
@@ -805,27 +826,27 @@ def reconnect(self):
...
+            # SSL is only supported when SSLContext is available (implies Python >= 2.7.9 or >= 3.2)
 
-        if self._transport == "websockets":
-            if self._tls_ca_certs is not None:
-                sock = WebsocketWrapper(sock, self._host, self._port, True)
+            verify_host = not self._tls_insecure
+            try:
+                # Try with server_hostname, even it's not supported in certain scenarios
+                sock = self._ssl_context.wrap_socket(sock, server_hostname=self._host)
+            except ValueError:
```
PierreF: ssl.CertificateError get caught by the except ValueError. Which means that if certificate validation failed (and server_hostname is supported) we will try to re-wrap the socket (which fail).

We probably should add a except ssl.CertificateError:\n raise juste before.

Nzbuu: :( Thanks for checking this out!

Nzbuu: Should be fixed now.

Looking at the ssl package docs, ssl.CertificateError should be the only error raised by the ssl code that isn't derived from OSError.

Lokesh-K-Haralakatta:
When can we get the SNI support released with paho.mqtt.python pkg? We need this feature to support client side certificates in our connections to IoT Platform.

Nzbuu:
@Lokesh-K-Haralakatta , might be better to ask at on the Paho-dev mailing list: https://dev.eclipse.org/mailman/listinfo/paho-dev

PR #93, opened 2016-07-08
Add Travis build configuration and Fix Python 3 tests

Nzbuu:
Looks like the Python 3 tests are currently broken


Nzbuu:
Python 3 tests have now been fixed, but `make test` only runs tests in current Python environment.

Travis runs tests on all supported Python interpreters, e.g. https://travis-ci.org/eclipse/paho.mqtt.python/builds/143852098.

Nzbuu:
@ralight, can I suggest that, when you have time, you prioritise this PR first, so that the other PRs can benefit too? Thanks

ralight:
Right, thanks for this there's a good amount of work gone into it. I think the tradeoff of having it tested under travis is worth the disadvantage that we no longer get Python 2 and 3 together.

Nzbuu:
Thanks for merging this. 

As mentioned in https://github.com/eclipse/paho.mqtt.python/issues/94, I don't think that `make test` didn't actually ran the tests in both python environments.

ralight:
It worked fine in my setup :) It was clear from what you said in #94 that it wasn't the best though.

In case you're not on the list, can I draw your attention to this post? http://dev.eclipse.org/mhonarc/lists/paho-dev/msg03661.html

PR #95, opened 2016-07-09
Add functionality to support standard Python logging

Nzbuu:
Implement support for logging using the standard logging package (see [PEP 282](https://www.python.org/dev/peps/pep-0282/)).

This has been implemented in a way that preserves existing functionality and behaviour exactly. I think that it might be better in the long run to modify the existing functionality to mirror the logging package more directly. However, this might break backwards compatibility.

In particular, I would:

* change value of the `MQTT_LOG_*` constants to equal the associated constant from the logging package, i.e. `MQTT_LOG_DEBUG = logging.DEBUG`
* delete MQTT_LOG_NOTICE, since it has no direct equivalent and is not currently used

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -268,6 +268,38 @@ Do not use this function in a real system. Setting value to True means there is 
 
 Must be called before ``connect*)``.
 
+enable_logger()
+'''''''''''''''
+
+::
+
+    enable_logger(logger)
+
+Enable logging using the standard python logging package (See PEP 282). This may be used at the same time as the ``on_log`` callback method.
+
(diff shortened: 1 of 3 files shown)


ralight:
Just to say, keep the requests coming. I'm not ignoring them, just having to work on other things at the moment.

ralight:
Unless anybody is using the actual numerical values for the logging constants then I can't think that there should be any real problem with backwards compatibility. Your proposals sound good to me.

ralight:
Oh, and thanks for another good pull request :)

PR #97, opened 2016-07-17
Rename examples to include relevant module name

Nzbuu:
This makes it clearer what functionality each example is aiming to demonstrate


Nzbuu:
Sorry, wrong target!

PR #98, opened 2016-07-17
Rename examples to include relevant module name

Nzbuu:
This makes it clearer what functionality each example is aiming to demonstrate


Nzbuu:
I've also updated the examples to use common code to ensure paho is on the path

ralight:
Thanks, I agree this makes things clearer. Sorry for the delay dealing with it.

PR #99, opened 2016-07-19
Introduce TLS server name indication (SNI) support

BrianEricsonExosite:
Update tls_set to add optional parameter "server_hostname", allowing
clients to specify server name indication (SNI) when connecting via TLS.

Replace ssl.wrap_socket with ssl.SSLSocket as ssl.wrap_socket does not
support passing the server_hostname (and its recommended alternative,
SSLContext.wrap_socket, does not support passing any of the optional
parameters already being passed even if it supports server_hostname).

Signed-off-by: Brian Ericson <brianericson@exosite.com>

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -523,6 +523,7 @@ class Client(object):
         self._tls_ca_certs = None
         self._tls_cert_reqs = None
         self._tls_ciphers = None
+        self._tls_server_hostname = None
         self._tls_version = tls_version
         self._tls_insecure = False
         # No default callbacks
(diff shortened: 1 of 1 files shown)


Nzbuu:
I think that the SSLContext version is still the recommended approach. I have implemented it in PR https://github.com/eclipse/paho.mqtt.python/pull/92, and it still allows access to all of the previous SSL options and has better encapsulation.

I oppose adding even more SSL properties to the client class since this is poor encapsulation, IMO. There's also an issue that the added attribute duplicates information from the `_host` attribute. However, @ralight might prefer this since it's more in keeping with the current implementation.

If this version is selected then I think that it will need some work to ensure that it doesn't break compatibility with Python versions and SSL libraries that don't support SNI.

ralight:
@BrianEricsonExosite thanks for your work with this, I think the solution in #92 is the better approach though.

PR #104, opened 2016-08-02
Added validation of the host name in mqtt.client.connect()

dreamiurg:
Signed-off-by: Dmitry Guyvoronsky <dmitry.guyvoronsky@gmail.com>

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -706,8 +706,19 @@ class Client(object):
         broker. If no other messages are being exchanged, this controls the
         rate at which the client will send ping messages to the broker.
         """
-        if host is None or len(host) == 0:
-            raise ValueError('Invalid host.')
+        import re
+
+        if host is None or len(host) == 0 or len(host) > 255:
+            raise ValueError('Invalid host "{0}": length should be between 1 and 255 characters.'.format(host))
+        if host[-1] == ".":
+            host = host[:-1] # strip exactly one dot from the right, if present
+
+        RE_VALID_IP_ADDRESS = re.compile("^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$", re.IGNORECASE)
+        RE_VALID_HOSTNAME = re.compile("^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$", re.IGNORECASE)
+
+        if not RE_VALID_IP_ADDRESS.match(host) and not RE_VALID_HOSTNAME.match(host):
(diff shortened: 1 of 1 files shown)


Nzbuu:
Is it necessary to hard-code this logic here? What about IPv6 addresses?

Also, does it add anything over just waiting for `socket.create_connection` to fail instead?

dreamiurg:
@Nzbuu, generally I would I prefer EAFP over LBYL (look-before-you-leap). However, in this case i thought that host name validation would give user more detailed error message. We've seen users passing incorrect hostname to the `connect()`, e.g `mqtt.client.connect("https://mqtt-endpoint.example.com", ...)` that lead to  `socket.getaddrinfo()` failing with pretty generic
`socket.gaierror: [Errno -2] Name or service not known` and caused some confusion.

What do you think about validating host name before trying to connect to it?

Nzbuu:
@dreamiurg, if your objective is just to throw a better exception, then my preference would be to do it in the `catch` block associated with opening the socket. As far as I am concerned, it's unnecessary to pre-check the host name, when `socket.create_connection` should be responsible for that. Especially considering that the checks that you have added are almost certainly not the same, for example IPv6 addresses are forbidden by this code and the `socket` package apply different constraints on the host name.

src/paho/mqtt/client.py
```diff
@@ -706,8 +706,19 @@ def connect_async(self, host, port=1883, keepalive=60, bind_address=""):
         broker. If no other messages are being exchanged, this controls the
         rate at which the client will send ping messages to the broker.
         """
-        if host is None or len(host) == 0:
-            raise ValueError('Invalid host.')
+        import re
```
Nzbuu: Would it be better to add this in the global imports? A user will always call this method at least once.

src/paho/mqtt/client.py
```diff
@@ -706,8 +706,19 @@ def connect_async(self, host, port=1883, keepalive=60, bind_address=""):
...
+        if host is None or len(host) == 0 or len(host) > 255:
+            raise ValueError('Invalid host "{0}": length should be between 1 and 255 characters.'.format(host))
+        if host[-1] == ".":
+            host = host[:-1] # strip exactly one dot from the right, if present
+
+        RE_VALID_IP_ADDRESS = re.compile("^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$", re.IGNORECASE)
+        RE_VALID_HOSTNAME = re.compile("^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$", re.IGNORECASE)
+
+        if not RE_VALID_IP_ADDRESS.match(host) and not RE_VALID_HOSTNAME.match(host):
+            raise ValueError('Invalid host "{0}": does not appear to be neither IP address nor hostname.'.format(host))
```
Nzbuu: This is very hard to test. I think this would be better extracted as a method or function.

dreamiurg:
After thinking about this for some more time, I think that pre-validation is not necessary. I'll explore other ways to give users more detailed error information instead.

PR #105, opened 2016-08-10
Fix issue with WebSocket connection

PierreF:
Avoid WebSocket handshake to hang forever. This is very similar to #90 : if due to some network condition the TCP socket hang, the WebSocket handshake will also hang.
Like for #90, it use keepalive as timeout.

If WebSocket handshake fail, loop_forever will no longer terminate but retry handshake (like it does for any other network errors). This especially occur if you have a proxy before MQTT broker (for example an HAproxy, I think any proxy that accept TCP connection while broker is down will cause the issue) but your MQTT broker is down.

To reproduce, I use a mqtt client:
```
#myclient.py
from paho.mqtt.client import Client

client = Client(transport='websockets')
client.connect_async('localhost', 8080)
client.loop_forever()
```
With a Mosquitto server with the following configuration:
```
#myconfig.conf
listener 18080
protocol websockets
```
And a proxy that accept connection. I use socat:
```
socat TCP-LISTEN:8080,fork,reuseaddr TCP:127.0.0.1:18080
```

The following step produce the error:
* Start mosquitto ```./src/mosquitto -v -c myconfig.conf``` and the socat proxy
* Start the client, it will connect correctly
* Stop the mosquitto server
* When client will retry to connect, the will crash with the following error

The errors is:
```
  File "myclient.py", line 5, in <module>
    client.loop_forever()
  File "/home/pierref/dev/ext/paho.mqtt.python/src/paho/mqtt/client.py", line 1407, in loop_forever
    self.reconnect()
  File "/home/pierref/dev/ext/paho.mqtt.python/src/paho/mqtt/client.py", line 833, in reconnect
    sock = WebsocketWrapper(sock, self._host, self._port, False)
  File "/home/pierref/dev/ext/paho.mqtt.python/src/paho/mqtt/client.py", line 2663, in __init__
    self._do_handshake()
  File "/home/pierref/dev/ext/paho.mqtt.python/src/paho/mqtt/client.py", line 2734, in _do_handshake
    raise ValueError("WebSocket handshake error")
ValueError: WebSocket handshake error

```

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -143,6 +143,11 @@ if sys.version_info[0] < 3:
 else:
     sockpair_data = b"0"
 
+
+class WebsocketConnectionError(ValueError):
+    pass
+
+
 def error_string(mqtt_errno):
     """Return the error string associated with an mqtt error number."""
     if mqtt_errno == MQTT_ERR_SUCCESS:
@@ -839,8 +844,10 @@ class Client(object):
 
         if self._transport == "websockets":
             if self._tls_ca_certs is not None:
+                self._ssl.settimeout(self._keepalive)
                 self._ssl = WebsocketWrapper(self._ssl, self._host, self._port, True)
             else:
+                sock.settimeout(self._keepalive)
                 sock = WebsocketWrapper(sock, self._host, self._port, False)
 
         self._sock = sock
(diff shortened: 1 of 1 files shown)


PierreF:
Can you have a look to this PR ?

ralight:
Does this still require #90?

PierreF:
Yes, this PR only fix connection hang for websocket part, not for SSL part which is done by #90

Nzbuu:
I think this change will cause a merge conflict with the changes in https://github.com/eclipse/paho.mqtt.python/pull/84,  which have already been merged into develop.

ralight:
Thanks, this looks fine. The merge conflict doesn't look a big deal.

PierreF, could I draw your attention to this post if you've not seen it on the list? http://dev.eclipse.org/mhonarc/lists/paho-dev/msg03661.html

PR #110, opened 2016-09-10
update setup.py to use setuptools and fall back to distutils when unavailable

fmoor:
also removed tab characters and replaced them with spaces, because tabs do not render the same in all editors.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,14 +2,19 @@ import sys
 sys.path.insert(0, 'src')
 from paho.mqtt import __version__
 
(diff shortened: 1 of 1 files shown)


fmoor:
It looks like the travis script is broken...

ralight:
Thanks for this, but the same changes have already been made in the develop branch.

fmoor:
I guess I should have started there...

ralight:
:)

Even if you had, I only merged it the other day so there may well have been an overlap.

Issue #107, opened 2016-08-24
Max mqtt payload size


MarcoFierimonte:
Hello,

Is there any limit on the message size that a Client Mqtt Paho can receive? Furthermore can this payload size depends from Operating System or programming language (java, python, etc..) ?

Thanks.
Marco

miketran78727:
@MarcoFierimonte MQTT allows applications to send Control Packets of size up to 268,435,455 (256 MB) http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/errata01/os/mqtt-v3.1.1-errata01-os-complete.html#_Toc442180836 

However, the limit can be imposed by the broker for practical purposes.  For example, IBM MessageSight has a concept of Endpoint (port) which allows the administrator to set maximum message size. https://www.ibm.com/support/knowledgecenter/SSCGGQ_1.2.0/com.ibm.ism.doc/Administering/ad00360_.html

ralight:
As Mike says, this is more likely to be enforced at the server side than the client side.

MarcoFierimonte:
@miketran78727 & @ralight thanks for your replies. I read MQTT specs but I checked different behaviour of Paho Clients on SO like: Raspbian, Windows and CentOS. On Rasbian the client Paho (Python) can't receive more than 2 KB of message payload. Instead on CentOS the maximum size is 4 KB.  It makes sense? 
In my project I implemented a MQTT broker using a library called Netty and different Paho clients communicate with it. It works well with Java Paho clients but not with Python clients, they give me the error "ERRNO 11" if payload size is big (around 2-4 KB).

Thanks.
Marco

ralight:
I've tested this code on raspbian and it receives a 200kB message just fine. Could you test at your end?

    import paho.mqtt.client as mqtt

    def on_message(mqttc, obj, msg):
        print(msg.topic+" "+str(len(msg.payload)))

    mqttc = mqtt.Client()
    mqttc.on_message = on_message
    mqttc.connect("test.mosquitto.org", 1883, 60)
    mqttc.subscribe("size", 0)
    mqttc.loop_forever()

MarcoFierimonte:
If you tested it and works the problem could be my Java Netty Broker. I am trying to find the bug. Thank you for your help @ralight  :)

Marco

MarcoFierimonte:
I missed a important part @ralight . I am using the websocket protocol to communicate. Maybe this is the problem, 2KB is the client paho on websocket limit? :)

[...]

PR #114, opened 2016-09-27
[114] Fix large payload over Websockets

PierreF:
Related to issue #107, when payload is too big, Websocket connection does not work. Issue is that  [_buffrered_read](https://github.com/eclipse/paho.mqtt.python/blob/33b7c95df3df330ab2ffb7b325e94164db1159fd/src/paho/mqtt/client.py#L2760) assume that socket.recv(wanted_size) will return data of size wanted_size, which is not always true, especially when wanted_size is larged than the TCP window size.

This PR will fix this issue, large payload (tested up to 100 Mb) works correctly.

To reproduce the issue:

Mosquitto server:

```
$ cat mosquitto.conf
listener 1884
listener 1885
protocol websockets
$ mosquitto -c mosquitto.conf
```

Create data to send/recv:

```
$ dd if=/dev/urandom of=data.bin bs=1024 count=10240
```

Start the client:

```
$ cat client.py
import paho.mqtt.client

def on_connect(client, ud, flags, rc):
    client.subscribe('topic')

def on_message(client, ud, message):
    print('Got the message!')
    data = open('data.bin', 'rb').read()
    if message.payload != data:
        print('Message mismatch')
    client.disconnect()

client = paho.mqtt.client.Client(transport='websockets')
client.on_message = on_message
client.connect('127.0.0.1', 1885)
client.loop_forever()
$ python client.py
```

Send the message:

```
$ mosquitto_pub -p 1884 -f data.bin -t topic
```

Without the PR, the following error occur:

```
  File "client.py", line 20, in <module>
    client.loop_forever()
  File "/home/pierref/dev/ext/github.com/eclipse/paho.mqtt.python/src/paho/mqtt/client.py", line 1354, in loop_forever
    rc = self.loop(timeout, max_packets)
  File "/home/pierref/dev/ext/github.com/eclipse/paho.mqtt.python/src/paho/mqtt/client.py", line 876, in loop
    rc = self.loop_read(max_packets)
  File "/home/pierref/dev/ext/github.com/eclipse/paho.mqtt.python/src/paho/mqtt/client.py", line 1159, in loop_read
    rc = self._packet_read()
  File "/home/pierref/dev/ext/github.com/eclipse/paho.mqtt.python/src/paho/mqtt/client.py", line 1708, in _packet_read
    data = self._sock.recv(self._in_packet['to_process'])
  File "/home/pierref/dev/ext/github.com/eclipse/paho.mqtt.python/src/paho/mqtt/client.py", line 2881, in recv
    return self._recv_impl(length)
  File "/home/pierref/dev/ext/github.com/eclipse/paho.mqtt.python/src/paho/mqtt/client.py", line 2825, in _recv_impl
    payload[index] ^= mask_key[index % 4]
IndexError: bytearray index out of range
```

A side note, receiving is slow on websocket, but this seems to be due to Mosquitto. Receiving that 10Mb payload took about 20 seconds when using mosquitto as MQTT broker. With mosca it took less than 1 second.
Mosa run with:

```
$ mosca -v --port 1884 --http-port 1885
```

It seems that Mosquitto is sending a packet of 64kb every 100ms. The 100ms is so fixed that is really looks like a sleep(100ms) somewhere

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -2760,14 +2760,18 @@ class WebsocketWrapper:
     def _buffered_read(self, length):
 
         # try to recv and strore needed bytes
-        if self._readbuffer_head + length > len(self._readbuffer):
+        wanted_bytes = length - (len(self._readbuffer) - self._readbuffer_head)
+        if wanted_bytes > 0:
 
-            data = self._socket.recv(self._readbuffer_head + length - len(self._readbuffer))
+            data = self._socket.recv(wanted_bytes)
 
             if not data:
                 raise socket.error(errno.ECONNABORTED, 0)
             else:
                 self._readbuffer.extend(data)
+
+            if len(data) < wanted_bytes:
+                raise socket.error(errno.EAGAIN, 0)
 
         self._readbuffer_head += length
         return self._readbuffer[self._readbuffer_head-length:self._readbuffer_head]


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=114

PierreF:
There had already issue about the slowness of WebSocket on Mosquitto : https://github.com/eclipse/mosquitto/issues/23 & https://github.com/eclipse/mosquitto/issues/185.

Which means that it's not the client that misbehave and cause slowness.

minghuadev:
Hi, Would you try to apply the change suggested by #185 and see how long it takes to transfer the same 10M data? Thanks.

PR #124, opened 2016-11-08
Fix for bug #3

jjasper:
This is the fix for the issue where connection attempts can block forever during the TLS handshake on blocking sockets set with no timeout. Thanks to Pierre Fersing for the keepalive substitution for timeout.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -829,7 +829,10 @@ class Client(object):
                 ca_certs=self._tls_ca_certs,
                 cert_reqs=self._tls_cert_reqs,
                 ssl_version=self._tls_version,
-                ciphers=self._tls_ciphers)
+                ciphers=self._tls_ciphers,
+                do_handshake_on_connect=False)
+            self._ssl.settimeout(self._keepalive)
+            self._ssl.do_handshake()
 
             if self._tls_insecure is False:
                 if sys.version_info[0] < 3 or (sys.version_info[0] == 3 and sys.version_info[1] < 2):


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=3

src/paho/mqtt/client.py
```diff
@@ -829,7 +829,10 @@ def reconnect(self):
                 ca_certs=self._tls_ca_certs,
                 cert_reqs=self._tls_cert_reqs,
                 ssl_version=self._tls_version,
-                ciphers=self._tls_ciphers)
+                ciphers=self._tls_ciphers,
+                do_handshake_on_connect=False)
```
Nzbuu: Is this input used?

PierreF: Yes, as it avoid doing the SSL handshake immediately when wrap_socket() is called.

This allow to set a timeout before doing SSL handshake so that the handshake honor the timeout. Without this, with a crappy connection the SSL handshake could hang a long time (and I think I've encounter case where "long time" was days :()

Nzbuu: Sorry. My mistake. I should have expanded the diff to see more context.

PierreF:
@jjasper the IP validation fail, I think it's because you committed with an email that didn't signed the CLA. Could you sign the CLA with this email ? Thanks !

jjasper:
Done, apologies for the delay. Anyone know how to kick the pull request process off again?

PierreF:
I've also missed, but you didn't Signed-Off your commit. Since you will need to update the commit this will re-trigger the ip-validation check.

PierreF:
@jjasper did you see my remark about the missing Signed-Off ?

jjasper:
@PierreF What do you mean by "update commit"?

PierreF:
You can update a commit with --amend and push it with --force to overwrite an existing commit.
If you feel more comfortable with creating a new PR, it's also fine.

In any case, don't forget the --signoff when you commit.
Thanks in advance.

jjasper:
@PierreF trying to do this on an ipad through github is not going well. I may have to wait until I get home tomorrow and have access to my laptop.

PierreF:
Yes, I don't know if you can update commit from github

PierreF:
By looking a bit more to commit generated from github, I'm pretty sure it can't works with ip-validation bot: commit don't from github interface use "something@users.noreply.github.com" a email address.

I think the only solution is to do this from your laptop.

Please remove the Deleteme commit as they will probably interfere with the ip-validation bot.
Thanks in advance.

Issue #126, opened 2016-11-22
UnboundLocalError when websocket connection is closed immediately


michaelboulton:
When a websocket connection is closed immediately I get this error:

```
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 754, in run 
    self.__target(*self.__args, **self.__kwargs)
  File "/home/michael/.virtualenvs/dev/local/lib/python2.7/site-packages/paho/mqtt/client.py", line 2557, in _thread_main
    self.loop_forever(retry_first_connection=True)
  File "/home/michael/.virtualenvs/dev/local/lib/python2.7/site-packages/paho/mqtt/client.py", line 1382, in loop_forever
    rc = self.loop(timeout, max_packets)
  File "/home/michael/.virtualenvs/dev/local/lib/python2.7/site-packages/paho/mqtt/client.py", line 901, in loop
    rc = self.loop_read(max_packets)
  File "/home/michael/.virtualenvs/dev/local/lib/python2.7/site-packages/paho/mqtt/client.py", line 1181, in loop_read
    rc = self._packet_read()
  File "/home/michael/.virtualenvs/dev/local/lib/python2.7/site-packages/paho/mqtt/client.py", line 1678, in _packet_read
    command = self._ssl.read(1)
  File "/home/michael/.virtualenvs/dev/local/lib/python2.7/site-packages/paho/mqtt/client.py", line 2906, in read
    return self._recv_impl(length)
  File "/home/michael/.virtualenvs/dev/local/lib/python2.7/site-packages/paho/mqtt/client.py", line 2850, in _recv_impl
    frame = self._create_frame(WebsocketWrapper.OPCODE_CONNCLOSE, payload, 0)
UnboundLocalError: local variable 'payload' referenced before assignment
```

It's trivially fixed by adding `payload = ""` to the top of `WebsocketWrapper._recv_impl` - I can make a pull request but it's only a 1 line change!

PierreF:
Issue fixed by pr #130.

PR #130, opened 2016-12-19
Fix uninitialized variable in Websocket _recv_impl

PierreF:
This is the fix for #126.

It use bytearray() as initial value for payload (and not empty string  "") since payload is normally a bytearray().

As I didn't known how to produce OPCODE_CONNCLOSE using Mosquitto, I've used a dummy tornado Websocket server that just close Websocket:
```
import tornado.websocket
import tornado.ioloop
class EchoWebSocket(tornado.websocket.WebSocketHandler):
    def open(self):
        self.close()
    def on_message(self, *args, **kwargs):
        pass

def make_app():
    return tornado.web.Application([
        (r"/mqtt", EchoWebSocket),
    ])

app = make_app()
app.listen(8888)
tornado.ioloop.IOLoop.current().start()
```

Then a client connection to this server fail with the UnboundLocalError & stack from #126 :+1: 
```
>>> import paho.mqtt.client
>>> cl = paho.mqtt.client.Client(transport='websockets')
>>> cl.connect('127.0.0.1', 8888)
>>> cl.loop_forever()
[...]
  File "/home/pierref/dev/ext/github.com/eclipse/paho.mqtt.python/src/paho/mqtt/client.py", line 2893, in _recv_impl
    frame = self._create_frame(WebsocketWrapper.OPCODE_CONNCLOSE, payload, 0)
UnboundLocalError: local variable 'payload' referenced before assignment
```

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -2836,6 +2836,7 @@ class WebsocketWrapper:
             self._readbuffer_head = 0
 
             result = None
+            payload = bytearray()
 
             chunk_startindex = self._payload_head
             chunk_endindex = self._payload_head + length


swails:
Something looks off.  The `fixes` branch doesn't have a `.travis.yml` file (which is why the CI is failing), but `develop` does.

Furthermore, GitHub says `develop` is 52 commits ahead of master and 3 behind, while `fixes` is 8 commits ahead.

What's the justification for splitting `develop` and `fixes`?  It seems (to me) to be causing more confusion than clarity.  And what's getting pushed where and why?

ralight:
@swails The model is a very simplified version of http://nvie.com/posts/a-successful-git-branching-model/

`master` is always stable. Bug fixes go into the `fixes` branch, new features go into `develop`. When a new release is made, if it is going to be a bugfix release (1.2.x) then the changes from the `fixes` branch are merged into `master` and the release is tagged. If it is a feature release (1.x.0 or x.0.0) then everything from `develop` and `fixes` gets merged into `master`.

Does that make sense?

Nzbuu:
I think the only problem is that the fixes aren't merged into `develop` very often.

swails:
> I think the only problem is that the fixes aren't merged into develop very often.

That's exactly what I was afraid of.  I'd like to see anything that goes into `fixes` also go into `develop` (with little to no delay) -- what sense does it make to have a stable development branch that has known bugs fixed in another branch based off yet another branch?

The reference for the branching model also seems to suggest that bug fixes get quickly merged (or cherry-picked) over to `develop` as well.

swails:
@ralight -- yes, that model mostly makes sense, but the process of making a feature release seems slightly suboptimal to me.  What I'm more familiar with is merging `fixes` to `master` to make a bugfix release, and cutting a release branch from `develop`.  And bug fixes go into `develop` and back-ported to `fixes` where appropriate.

Merging both `develop` and `fixes` into `master` to make a new feature release means you end up with a release code base that nobody has ever seen, used, or tested.

Nzbuu:
I think in classic gitflow, both master and develop would be updated with each bugfix immediately when it's approved, and you would use tags on master to denote stable releases.

swails:
> I think in classic gitflow, both master and develop would be updated with each bugfix immediately when it's approved, and you would use tags on master to denote stable releases.

Agreed.  So any bugfix PR should come in a pair unless it applies only to develop, I think...

Was this bug already fixed in develop?

PierreF:
See #140 for discussion about branch model.

Issue #131, opened 2016-12-21
Library doesn't pay attention to the _ssl.pending() bytes 


olgierdh:
Hi,

My name is Olgierd Humeńczuk and I work for Xively@LogMeIn. We’ve encountered some problems while using your Paho Python implementation. The problem occurs when receiving retained messages after connecting to a broker using an SSL socket. It seems that after receiving data on the wire the SSL library keeps part of the data within its internal buffer and Paho Python does not check if this buffer is not empty. This results in a situation when the first message is received and parsed properly but the next message is kept within this SSL internal buffer. After adding the self._ssl.pending() > 0 condition to the paho’s loop function everything works as expecte.  Without this change the Paho Python client won’t receive all of the messages properly.

Do you know anything about that issue ? Perhaps we are not using your library properly?

This is how we’re using the client:

we set up paho by creating an instance and calling tls_set (in order to use SSL sockets)
we create a separate thread for executing the Paho client
we are calling loop( timeout=1.0 ) periodically from the worker thread to keep processing without blocking the thread

Our code of xiPy library and the described fix is available on github at https://github.com/xively/xiPy/pull/5/files

Best Regards
Olgierd Humeńczuk

PierreF:
Hi,

thanks for the report. There is indeed a bug in current implementation. We expected select.select() to return the SSL socket if some data are ready to be read... but it's not always true as said the [Python doc](https://docs.python.org/3/library/ssl.html#notes-on-non-blocking-sockets):
> a SSL socket may still have data available for reading without select() being aware of it.

The documentation suggest doing a SSLSocket.recv() (SSLSocket.read() should work the same), but I don't see how to do this with current architecture.

Your solution seems the best one.

Could you submit a PR with your changes ?

olgierdh:
Sure, I will create one.

PierreF:
PR #132 merged

PR #132, opened 2016-12-21
Fixes issue #131. 

olgierdh:
Makes paho.mqtt.python aware of the data pending in the SSL library. This makes the library to continue parsing messages that has been received and stored within SSL's internal buffer.

Signed-off-by: olgierdh <olgierd.humenczuk@logmein.com>


positron96:
It probably breaks websockets+TLS implementation: #182 . Could you please check?

PR #136, opened 2017-01-03
Fixes some pylint issues and improves data access in dictionaries.

moisesguimaraes:
Signed-off-by: Moisés Guimarães <guimaraes.moises@gmail.com>

### DIFF
--- a/src/paho/mqtt/publish.py
+++ b/src/paho/mqtt/publish.py
@@ -31,7 +31,7 @@ def _do_publish(client):
     if isinstance(message, dict):
         client.publish(**message)
     elif isinstance(message, tuple):
-        client.publish(message[0], message[1], message[2], message[3])
+        client.publish(*message)
     else:
         raise ValueError('message must be a dict or a tuple')
 
--- a/src/paho/mqtt/publish.py
+++ b/src/paho/mqtt/publish.py
@@ -29,10 +29,7 @@ def _do_publish(client):
     message = client._userdata.pop()
 
     if isinstance(message, dict):
-        client.publish(message['topic'],
-                       message.get('payload'),
-                       message.get('qos', 0),
-                       message.get('retain', False))
+        client.publish(**message)
     elif isinstance(message, tuple):
         client.publish(message[0], message[1], message[2], message[3])
     else:
--- a/src/paho/mqtt/publish.py
+++ b/src/paho/mqtt/publish.py
@@ -25,6 +25,7 @@ import paho.mqtt as mqtt
 
 def _do_publish(client):
     """Internal function"""
+
     message = client._userdata.pop()
 
     if isinstance(message, dict):
@@ -40,6 +41,8 @@ def _do_publish(client):
 
 def _on_connect(client, userdata, flags, rc):
     """Internal callback"""
+    #pylint: disable=invalid-name, unused-argument
+
     if rc == 0:
         _do_publish(client)
     else:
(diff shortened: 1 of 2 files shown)


src/paho/mqtt/publish.py
```diff
@@ -19,53 +19,45 @@
 broker, then disconnect and nothing else is required.
 """
 
+import ssl
```
PierreF: Why did you add an import ssl ?

src/paho/mqtt/subscribe.py
```diff
@@ -19,53 +19,53 @@
 you to pass a callback for processing of messages.
 """
 
+import ssl
```
PierreF: ssl is not used. We should just remove the import.

src/paho/mqtt/publish.py
moisesguimaraes: You're right, I was working on master branch, so I needed to import it. Now it is not needed anymore.

src/paho/mqtt/subscribe.py
moisesguimaraes: done

Nzbuu: Have you pushed this change yet?

src/paho/mqtt/subscribe.py
```diff
@@ -124,40 +124,25 @@ def callback(callback, topics, qos=0, userdata=None, hostname="localhost",
...
     client.on_connect = _on_connect
 
     if auth is not None:
-        username = auth['username']
-        try:
-            password = auth['password']
-        except KeyError:
-            password = None
-        client.username_pw_set(username, password)
+        client.username_pw_set(auth['username'],
```
Nzbuu: What about the following?

``
client.username_pw_set(**auth)
```

File: src/paho/mqtt/subscribe.py (hunk 2 above)
Comment (moisesguimaraes) [inline src/paho/mqtt/subscribe.py]: I removed `import ssl` from `publish.py` and forgot to remove from `subscribe.py`.

Now it is really done!

File: src/paho/mqtt/subscribe.py (hunk 3 above)
Comment (moisesguimaraes) [inline src/paho/mqtt/subscribe.py]: I did the same with `will_set()`.

File: src/paho/mqtt/publish.py [hunk 4]
```diff
@@ -23,49 +23,45 @@
...
-            retain = m['retain']
-        except KeyError:
-            retain = False
-    elif type(m) is tuple:
-        (topic, payload, qos, retain) = m
+
+    message = client._userdata.pop()
+
+    if isinstance(message, dict):
+        client.publish(message['topic'],
```
moisesguimaraes: @Nzbuu,

Can I do the same here?

client.publish(**message)

If the dictionary has more keys than the function has parameters, we will get an error like this:

TypeError: publish() got an unexpected keyword argument 'foo'

PierreF: This is already true for the username_pw_set or will_set. I would say to keep consistent and do the same.

moisesguimaraes: done.

Nzbuu: @moisesguimaraes. By expanding the dictionary, all of the key/value pairs are passed to the publish method, not just those that are expected. Hence, it will throw an exception if there are unexpected inputs, which I think is the clearest behaviour.

src/paho/mqtt/subscribe.py
```diff
@@ -21,51 +21,50 @@
...
-def _on_message_simple(c, userdata, message):
+def _on_message_simple(client, userdata, message):
     """Internal callback"""
 
     if userdata['msg_count'] == 0:
         return
 
     # Don't process stale retained messages if 'retained' was false
-    if userdata['retained'] == False and message.retain == True:
+    if userdata['retained'] is False and message.retain is True:
```
PierreF: Yes, I agree.

PR #138, opened 2017-01-04
Socket changes

michaelboulton:
I noticed that there was a lot of unnecessary if/else depending on if an ssl socket or a normal socket was used when they behave basically the same in almost all contexts - this pull request just fixes this issue so that it doesn't need to keep checking whether it's an ssl socket or not

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -455,9 +455,7 @@ class Client(object):
       MQTT_LOG_ERR, and MQTT_LOG_DEBUG. The message itself is in buf.
 
     """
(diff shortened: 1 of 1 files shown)


PierreF:
Thanks for your PR, but I think most of your changes are already fixed in develop branch.
Do you see anything in your PR not already done in develop ?

michaelboulton:
I think there's only one tiny thing different than the develop branch where it passes the is_ssl into the websocket handler initialiser - I imagine this could potentially be used in future so I don't think it's worth making a new pull request for that. I'll close this

PR #139, opened 2017-01-04
Change logging to use python log facilities

michaelboulton:
Logging from the library currently requires writing a custom handler to handle the different levels, and there also isn't actually that much debug logging in there anyway.

This pull request makes the library use the built in python logging so that it can all be done automatically outside of the library (eg, reading a yaml file which has a dictionary that can be passed directly into dictconfig) and adds a bit more debug logging.

I also removed the logging of _userdata from all easy_log messages - This is purely because in my experience it never really seemed useful to log, but it can be added back in at the specific places where it was logged before if necessary.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -33,6 +33,7 @@ except:
 import struct
 import sys
 import threading
+import logging
 
 import time
 import uuid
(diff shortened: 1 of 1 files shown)


Nzbuu:
There have already been changes to allow standard Python logging in the `develop` branch. Can you review that and make a new pull request against the `develop` branch if necessary.

Nzbuu:
Also, these changes break backwards-compatibility, so is unlikely to be accepted.

michaelboulton:
Seems like it's all in the develop branch!

PR #157, opened 2017-03-05
Detect higest TLS versions since Python 3.6

dennisdegreef:
See https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS

Since Python 3.6, this constant can auto detect TLS version, and thus automatically choose TLSv1.2 or higher if applicable, instead of hardcoded TLSv1.0.

The reason behind this change is that most people won't call `set_tls` in implementation.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -641,6 +641,9 @@ class Client(object):
         # Create SSLContext object
         if tls_version is None:
             tls_version = ssl.PROTOCOL_TLSv1
+            # If the python version supports it, use highest TLS version automatically
+            if hasattr(ssl, "PROTOCOL_TLS"):
+                tls_version = ssl.PROTOCOL_TLS
         context = ssl.SSLContext(tls_version)
 
         # Configure context
--- a/README.rst
+++ b/README.rst
@@ -232,7 +232,7 @@ tls_set()
 ::
 
     tls_set(ca_certs=None, certfile=None, keyfile=None, cert_reqs=ssl.CERT_REQUIRED,
-        tls_version=ssl.PROTOCOL_TLSv1, ciphers=None)
+        tls_version=ssl.PROTOCOL_TLS, ciphers=None)
 
 Configure network encryption and authentication options. Enables SSL/TLS support.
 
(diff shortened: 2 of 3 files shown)


PierreF:
Few remarks:

* If I understand documentation correctly, this would allow SSLv3. We need to set OP_NO_SSLv3 (also OP_NO_SSLv2 ?) to avoid a possible less secure default.
* Your PR should be done against develop branch.
* Please update docstring to reflect the change in the default value

Also if not already done, you will need to sign the Eclipse CLA: https://www.eclipse.org/contribute/cla

Nzbuu:
I think that `ssl.PROTOCOL_TLS` is also in Python 3.5.3 and 2.7.13. So you have to do the detection via `hasattr` rather than by Python version number.

src/paho/mqtt/client.py
```diff
@@ -24,8 +24,14 @@
 HAVE_SSL = True
 try:
     import ssl
+    import sys
     cert_reqs = ssl.CERT_REQUIRED
-    tls_version = ssl.PROTOCOL_TLSv1
+
+    # Since 3.6 this constant selects highest protocol
+    if sys.hexversion > 0x03060000:
```
Nzbuu: This check is too specific since there's support in other Python version. Use `hasattr` instead.

src/paho/mqtt/client.py
```diff
@@ -24,8 +24,14 @@
 HAVE_SSL = True
 try:
```
Nzbuu: There's too much code in this try-catch. Everything after `import ssl` belongs in an `else` clause.

Nzbuu:
In case it wasn't clear. Thanks for making this pull request. It's definitely worthwhile. 👍

Nzbuu:
@PierreF , I think that SSLv2 and SSLv3 are disabled by default: https://docs.python.org/3/library/ssl.html#id5

Nzbuu:
@dennisdegreef , will you be able to update this pull request?

dennisdegreef:
@Nzbuu Done :+1: As far as I can tell, all feedback is processed.

PR #160, opened 2017-03-13
Run tests via tox

Nzbuu:
Also removed vagrantfile


Nzbuu:
I'm not sure that running `make test` via tox is a good idea, since it doesn't work on Windows

PR #161, opened 2017-03-14
Make tests run correctly on Windows

Nzbuu:
`make test` in `test/lib` should work on Windows, provided `make` is installed


Nzbuu:
@ralight , @PierreF , Do you think this can be merged?

Issue #163, opened 2017-03-15
Proposal: Drop support for Python 3.2


Nzbuu:
Python 3.2 is not supported by the standard Python packaging tools, such as pip, setuptools and virtualenv since Jan 2016. Can we drop support too?

The argument in [pypa-dev mailing list](https://groups.google.com/forum/#!msg/pypa-dev/Ef0PF2ZGAv0/hrO4BHkOBQAJ) was that there is minimal Python 3.2 usage in April 2015.

Perhaps more importantly, Python 3.2 was officially end-of-life in February 2016 (See https://www.python.org/dev/peps/pep-0392/)

See:

* https://pypi.python.org/pypi/pip
* https://pypi.python.org/pypi/setuptools
* https://pypi.python.org/pypi/virtualenv

PierreF:
One drawback is that Debian 7 (wheezy, currently oldstable) use Python 3.2 and is supported (by the LTS support) up to May 2018.
Is there any feature that fail to work on Python 3.2 ? You referenced this issue only from test and fixed them for Python 3.2.

I would say that if it don't require too much effort, we should try to keep support for Python 3.2.

swails:
Arguments against 3.2 support:

1. Security fixes ended over a year ago at least: https://www.python.org/dev/peps/pep-0392/
2. http://www.curiousefficiency.org/posts/2015/04/stop-supporting-python26.html

The first is a bit more worrying for MQTT, in my opinion, as it is inherently a protocol for communicating information to/from anywhere. The second is about 2.6 specifically, but the arguments still apply to 3.2.

swails:
If paho incidentally works in Python 3.2, that's fine(ish). But I think explicitly supporting a EOLed platform with known and publicly documented security vulnerabilities doesn't make sense. And I don't think it's worth even a nominal effort to keep supporting it. Just my opinion.

Nzbuu:
**TL:DR;** I think that my preference is to say that the develop branch is preparing for Paho 2.x since it may have breaking changes, and that Paho 2.x will only releases of Python that are not going to be EOL on the target release date.

[...]

PR #164, opened 2017-03-15
Start adding unit test suite in py.test

Nzbuu:
Here's the start of a unit test suite using py.test showing how it might work. Migrating the other tests will be much harder since they mock the MQTT server.

Note that this requires pip and pytest, which require extra logic to support Python 3.2 (see issue #163).


Nzbuu:
Does anyone have any comments on this? @PierreF ?

Makefile
```diff
@@ -32,6 +31,8 @@ clean-test: ## remove test and coverage artifacts
 	rm -fr htmlcov/
 
 test :
+    python ./setup.py develop
```
PierreF: This make the makefile invalid. Makefile require tab, not space.
Also this will (try to) install the package in *system* (/usr/local).

Makefile
```diff
@@ -32,6 +31,8 @@ clean-test: ## remove test and coverage artifacts
...
+    py.test
```
PierreF: This does not work on a fresh virtualenv. The setup.py develop won't install py.test.
A workaround is to run python setup.py test (which seems to works without requiring develop or any virtualenv)

Nzbuu: I think that I'll remove the setup step from the test target in the Makefiles. Otherwise it wouldn't be possible to test an officially installed version.

Makefile
Nzbuu: I think that I'll remove the setup step from the test target in the Makefiles

Makefile
```diff
@@ -1,18 +1,20 @@
...
-.PHONY : all clean clean-build clean-pyc clean-test install test upload
+.PHONY : all clean clean-build clean-pyc clean-test install test upload develop
 
 all :
 	python ./setup.py build
 
 install : all
 	python ./setup.py install --root=${DESTDIR}
 
+develop :
```
PierreF: Is this target needed ?

python setup.py develop will already install dependency for *running* paho-mqtt (currently nothing).
It only add dependency needed to run test, which is handled by python setup.py test.

Nzbuu: It's probably not required. Although `setup develop` will also ensure that paho is on the PYTHONPATH. I don't really mind if it's included. I develop on Windows, so I wouldn't use it.

Nzbuu: @PierreF , what would you prefer?

PierreF: I prefer removing it.
I think that a developer won't necessary known this make target but will probably known the python setup.py develop command and will run it directly.

Nzbuu: Done

Nzbuu:
I've removed the extra logic for Python 3.2 now (#163)

Nzbuu:
OK. I've added a candidate conversion of `test_01_con_discon_success` using py.test.

I'm pretty sure that there are better ways of implementing it, but this one works for now and shows that many of the existing tests can be converted reasonably directly.

tests/test_client.py
```diff
@@ -0,0 +1,49 @@
...
+            proto_name="MQIsdp", proto_ver=3)
+        fake_broker.expect_packet(connect_packet)
+
+        connack_packet = paho_test.gen_connack(rc=0)
+        fake_broker.send_packet(connack_packet)
+
+        disconnect_packet = paho_test.gen_disconnect()
+        fake_broker.expect_packet(disconnect_packet)
+
+        mqttc.loop_stop()
```
PierreF: Should we put the loop_stop in a finally clause ?
What I fear is that if some asserting fail before, the background thread will continue to run and may interfere with next test.

tests/testsupport/broker.py
```diff
@@ -0,0 +1,62 @@
...
+        if self._sock is not None:
+            self._sock.close()
+            self._sock = None
+
+    def expect_packet(self, packet_expected):
+        if self._conn is None:
+            raise ValueError('Connection is not open')
+
+        packet_in = self._conn.recv(len(packet_expected))
+        if not packet_in:
```
PierreF: Why this test return a RuntimeError ? A simple assert seems more adapted, since it's part of the test.

tests/testsupport/broker.py
```diff
@@ -0,0 +1,62 @@
...
+            raise RuntimeError('Connection was closed')
+        assert packet_in == packet_expected
```
PierreF: Currently if the assert fail, the message looks like:
```
self = <testsupport.broker.FakeBroker instance at 0x7f655b701d88>, packet_expected = '\x10#\x00\x06MQIsdp\x03\x02\x00<\x00\x1502-con-discon-success'

    def expect_packet(self, packet_expected):
[...]
>       assert packet_in == packet_expected
E       AssertionError
```
e.g. we don't have at all the packet_in value and expected packet is show un-decoded.

We should at least show the packet_in value. I think of two way to do that:
* always print() packet received/sent/expected. Since py.test capture stdout and only show it on error it should not generate noise and may help to understand failure since all network exchange will be logged.
* Add a assert message. Something like:
```
assert packet_in == packet_expected, "Expected = %s\nReceived = %s" % (decoded_expected, decoded_received)
```

tests/test_client.py
Nzbuu: Yes. That would probably be a good idea.

tests/testsupport/broker.py
Nzbuu: I had expected pytest rewrite the assertion: https://docs.pytest.org/en/latest/assert.html. I think that the issue is that the assertion is made in another module, rather than the test module, although pytest is imported there too.

I've re-written the test as follows:

```
            connect_packet = paho_test.gen_connect(
                "01-con-discon-success-fail", keepalive=60,  # Wrong client id
                proto_name="MQIsdp", proto_ver=3)
            packet_in = fake_broker.receive_packet(len(connect_packet))
            assert packet_in == connect_packet
```

And pytest now says:

```
>           assert packet_in == connect_packet
E           AssertionError: assert '\x10#\x00\x0...iscon-success' == '\x10(\x00\x06...-success-fail'
E             - \x10#\x00\x06MQIsdp\x03\x02\x00<\x00\x1501-con-discon-success
E             ?     ^                                  ^
E             + \x10(\x00\x06MQIsdp\x03\x02\x00<\x00\x1a01-con-discon-success-fail
E             ?     ^                                  ^                     +++++
```

I hope you'll agreed that this is much more useful than a simple `print`.

tests/testsupport/broker.py
Nzbuu: That's not really an assertion (i.e. thing that we are testing) rather an unexpected event.

This is just a first version. There needs to be **a lot** of refactoring to make paho more testable. I want to be using a fake socket for most of the tests in the future, but paho currently lacks the dependency injection functionality that would be required.

tests/test_client.py
Nzbuu: It would be nicer to use a context manager in the future though ...

tests/testsupport/broker.py
Nzbuu: I've moved all of the assertions to the test functions

Nzbuu:
@PierreF , I think that we should merge this PR and work on converting the rest of the tests in slower time.

PR #165, opened 2017-03-20
Make Travis tests run on `fixes` branch

Nzbuu:
Also fixes tests on Windows

Note that tests fail on Python 3.2 due to use of `u""` literal in test case (see #163).


Nzbuu:
Python 3.2 tests now passing by using `from __future__ import unicode_literals`.

PR #168, opened 2017-03-24
Use mutexes as context managers

michaelboulton:
The mutexes in the current version of the library are currently manually acquired and released when they are used. This makes the code harder to read and understand, and also produces some hard to track down bugs - I personally had an issue that I could not track down to do with on_disconnect but after changing the majority of the use of mutexes to use context mangers I found the problem went away.

This merge request changes _most_ of the use of mutexes to use them as context managers instead of manually acquiring and releasing them.

I've also added mutexes to the setters for on_connect, on_disconnect, etc. It seems that if the getters are protected by a mutex (such as in _handle_suback) then setting them should also be protected with a mutex, otherwise there could be a race condition a long the lines of:

1. User in main thread sets on_subscribe to be a function
2. Client acquires _callback_mutex, and self._on_subscribe is present
3. Before the function is actually called, user sets on_subscribe to be None
4. Client tries to call None, raises an exception

There are a few notes about this merge request which I think need to be finalised as well - I can change the merge request to fix any of these issues:

- Some of the blocks where mutexes are manually acquired/released have to be left in without some more significant changes to the code (or releasing and acquiring the mutexes again). For example, in publish() where the _out_message_mutex is released because _send_publish is called immediately afterwards, which in turn acquires the mutex. Changing the flow of the function might allow the use of a mutex with context manager instead but it might just make it unnecessarily complicated.

- Related to the above point, there are a couple of places (_handle_pubrel and _handle_pubrec) where there is a similar problem with mutexes being held inside the context manager, but I'm not sure that there is any way in which it would actually cause an issue. For example, in _handle_pubrec it acquires the _out_message_mutex then possibly returns the value of a call to _send_pubrel - For the duration of this call the _out_message_mutex is still being held, but it seems like none of the functions in the call tree from _send_pubrel will ever actually try to acquire this mutex. If there could be a situation I haven't seen where this could cause deadlock, I can change it back to the way it was handling the mutexes before my changes.

- loop_forever was changed slightly to use a function to check whether it should exit rather than having the mutex being acquired/released after the if statements. This was just for readability and easy of understanding the control flow.

-  _in_callback was also changed to be a mutex rather than just a True/False check. This does mean that multiple callbacks can't be called at once any more, but I'm not sure this could ever be the case? The only place this value is ever checked is in _packet_queue, where the value is only relevant if the client is being used synchronously anyway - it seems desirable to have a mutex in this situation so that use threads don't interfere.

- _state was changed to use a property like the callback functions to eliminate the need for using _state_mutex every time it was accessed. Maybe this could be applied to other variables as well?

This is a relatively simple _idea_ for a change to the library but the added issue of ensuring thread safety did end up making it necessarily complicated.


src/paho/mqtt/client.py
```diff
@@ -524,7 +524,7 @@ def __init__(self, client_id="", clean_session=True, userdata=None, protocol=MQT
         self._host = ""
         self._port = 1883
         self._bind_address = ""
-        self._in_callback = False
+        self._in_callback = threading.Lock()
```
Nzbuu: I don't know if this is supposed to be a lock. Rather just an indicator that a callback is being executed.

Nzbuu:
Thanks, this has been on my wish list for a while.

michaelboulton:
I have signed the ECA now - I completely forgot to do it before submitting the request!

Nzbuu:
There's a test failure on Python 2.6: https://travis-ci.org/eclipse/paho.mqtt.python/jobs/214721666

michaelboulton:
Pushed a fix for that now.

PierreF:
+1 on using context manager.

 > Some of the blocks where mutexes are manually acquired/released have to be left in without some more significant changes to the code (or releasing and acquiring the mutexes again). For example, in publish() where the _out_message_mutex is released because _send_publish is called immediately afterwards, which in turn acquires the mutex. Changing the flow of the function might allow the use of a mutex with context manager instead but it might just make it unnecessarily complicated.

If we could use context manager for all locks it would be better, but if not doable in a clean way we could stay with your current PR.
That being said, with RLock are issue goes away, no ? It there any downside to using RLock instead of unlocking them manually to re-acquire them later in calltree ?

 > Related to the above point, there are a couple of places (_handle_pubrel and _handle_pubrec) where there is a similar problem with mutexes being held inside the context manager, but I'm not sure that there is any way in which it would actually cause an issue. For example, in _handle_pubrec it acquires the _out_message_mutex then possibly returns the value of a call to _send_pubrel - For the duration of this call the _out_message_mutex is still being held, but it seems like none of the functions in the call tree from _send_pubrel will ever actually try to acquire this mutex. If there could be a situation I haven't seen where this could cause deadlock, I can change it back to the way it was handling the mutexes before my changes.

If it understand correctly, that means holding a lock a little longer that what is needed. If it does have visible performance impact that seems fine.

> _in_callback was also changed to be a mutex rather than just a True/False check. This does mean that multiple callbacks can't be called at once any more, but I'm not sure this could ever be the case? The only place this value is ever checked is in _packet_queue, where the value is only relevant if the client is being used synchronously anyway - it seems desirable to have a mutex in this situation so that use threads don't interfere.

I agree that if two callback could be called, this could create issue (when first callback finish, the _in_callback is reset to False). Using a lock seems better.

> _state was changed to use a property like the callback functions to eliminate the need for using _state_mutex every time it was accessed. Maybe this could be applied to other variables as well?

If other variable could be used as property it could be good... but as soon have you have some kind of "read-compute-write" pattern I'm not sure property will work. e.g.:
```
self._out_message_mutex.acquire()
if len(self._out_messages) >= max_size:
   exit()
self._out_messages.append(message)
```

In fact, I'm not sure we need a mutex around self._state :)



If you are going to work further on this PR, and if you have time to look at #177 that would be great. If not I would fix that after merging this PR.

michaelboulton:
> In fact, I'm not sure we need a mutex around self._state :)

I think you're right with this - iirc all the standard data structures in Python are thread safe? I don't think there could be any issues from removing it just looking at the code. Should I change this so there's no mutex around self._state?

> That being said, with RLock are issue goes away, no ? It there any downside to using RLock instead of unlocking them manually to re-acquire them later in calltree ?

I think in that specific case there should be no issue with using a RLock because you don't want external threads directly calling the various internal _send_x etc. methods anyway.


Working on #177  will ideally require writing a regression test for it, and I think I'd rather wait till https://github.com/eclipse/paho.mqtt.python/pull/164 is pulled in to fix that as I'm a lot more used to writing pytest tests.

michaelboulton:
I've changed out_message_mutex to an RLock, but I'm not sure about the _current_out_packet_mutex. This might cause some threading issues with the loop here https://github.com/ZoetropeLabs/paho.mqtt.python/blob/54b9b1ac424aaf401579cf7727cf20af562e4865/src/paho/mqtt/client.py#L1820 if the current_out_packet is changed mid-loop by the thread. I think this might be worth leaving as is for now.

Issue #162, opened 2017-03-15
can i set path of broker?


eduard739:
i want to set also the path of the broker(<host>/<path>), how can i do this?
examples: 
/iot.eclipse.org:80/ws
ec2-54-234--122.compute-1.amazonaws.com/eyJraWQiO...

PierreF:
You want to change the URL path when connecting to a broker over Websocket ?

Currently the "/mqtt" path is hardcoded, but it could indeed be a good feature to add.

eduard739:
yea. I found it in the WS wrapper class and changed it, thanks

PR #169, opened 2017-03-24
Arbitrary websocket headers

michaelboulton:
This adds functionality to send arbitrary authentication headers to the MQTT broker when an MQTT connection is used.

This makes it so that a function can be passed in to the Client constructor which takes the existing websocket headers and returns a dictionary corresponding to the new headers plus any extra required headers for authentication.

This also comes with an example function for users which shows how to generate headers which will authenticate with the AWS IoT platform from the AWS keys and the hash of the other headers, heavily based off the existing Amazon documentation.

There is also a minor fix to _recv_impl in the case where there is no payload.


src/paho/mqtt/client.py
```diff
@@ -2582,7 +2586,7 @@ class WebsocketWrapper:
     OPCODE_PING = 0x9
     OPCODE_PONG = 0xa
 
-    def __init__(self, socket, host, port, is_ssl):
+    def __init__(self, socket, host, port, get_auth_headers):
```
Nzbuu: You need to keep the `is_ssl` input

src/paho/mqtt/client.py
```diff
@@ -437,7 +437,8 @@ def on_connect(client, userdata, flags, rc):
 
     """
 
-    def __init__(self, client_id="", clean_session=True, userdata=None, protocol=MQTTv311, transport="tcp"):
+    def __init__(self, client_id="", clean_session=True, userdata=None,
+            protocol=MQTTv311, transport="tcp", get_auth_headers=None):
```
Nzbuu: You need to describe the `get_auth_headers` input.

Nzbuu:
I think that this functionality is a very good idea, but I'm concerned that the implementation has poor encapsulation. As a minimum, I think that I would prefer to see a separate method for configuring the websockets headers, rather than expanding the client constructor method. But that's just my view, @ralight , @PierreF .

michaelboulton:
I pushed a fix for the ssl argument error.

I agree that the constructor passing is probably not in the same 'spirit' as the other methods like `tls_set`. It's not a huge change to make to move it to a separate method (or something else entirely), I'll defer to whatever is deemed to be the best approach and I can implement that.

src/paho/mqtt/client.py
```diff
@@ -2723,6 +2737,8 @@ def _buffered_read(self, length):
 
     def _recv_impl(self, length):
 
+        payload = ""
```
PierreF: This is fixed in #130 (merged in fixes not develop). But #130 use bytearray(). IIRC this is the correct type for payload.
Could you either don't add this in the PR or use the same type.

PierreF:
I've quickly looked. Since other feature use additional method (like tls_set), I think it's better to keep consistent.

Does this need to be a method ? Why it's not a simple (ordered) dictionary that could be either directly set (an attribute - which could easily allow to read defaults) or set with .headers_set() ?

michaelboulton:
@PierreF the reason it was implemented like this was because of the AWS use case. You need to hash the date to generate the authentication headers and you can specify a timeout on the auth credentials, so if you generate them then try to use them later they could be expired.

Nzbuu:
The more I think about this, the more I feel that it should be a separate method (say, `ws_set_headers`). Then, it's much easier to include the necessary validation.

It would also be possible to handle different inputs, say either a `dict` (for static headers) or a `callable` (for dynamic ones).

PierreF:
It could be also good to be able to change the URL used in the request. Currently hardcoded to /mqtt, but #162 ask to parameterizable URL path.

michaelboulton:
I've now changed it so that there is a ws_set_options method which can be called to set the mqtt path or any extra headers. If the extra headers is a dict it is just appended to the existing headers, if it is a callable then the behaviour is the same as before where the existing headers are passed in and the result of the function is used as the new headers.

src/paho/mqtt/client.py
```diff
@@ -2651,29 +2676,41 @@ def __init__(self, socket, host, port, is_ssl):
...
+            "Sec-WebSocket-Key": sec_websocket_key,
+            "Sec-Websocket-Version": "13",
+            "Sec-Websocket-Protocol": "mqtt",
+        }
+
+        if extra_headers is not None:
+            if isinstance(extra_headers, dict):
+                websocket_headers.update(extra_headers)
+            elif callable(extra_headers):
+                websocket_headers = extra_headers(websocket_headers)
```
Nzbuu: Is it worth having an `else` clause here? If not, then I don't think you need to original `if is not None`.

michaelboulton: Good point, I copied that check into ws_set_options and forgot to remove it from _do_handshake

PierreF:
Small question about the order. Does AWS require header to be sorted in the request ? (for the the signature it has to be sorted, but currently we sort them in _do_handshake before sending them).
If yes, shouldn't prefer using a OrderedDict and keep the order that user callback provided ?

Also could you add this to README documentation. Remember to add the note that this method must be called before connect function (like tls_set).

Apart those 2 remark, LGTM

michaelboulton:
After looking at the documentation again, I think it doesn't need to be sorted in the request. In the [docs](http://docs.aws.amazon.com/general/latest/gr//sigv4-signed-request-examples.html#sig-v4-examples-get-auth-header) referring to the request:

> The request can include any headers, but MUST include "host", "x-amz-date", 
> and (for this scenario) "Authorization". "host" and "x-amz-date" must
> be included in the canonical_headers and signed_headers, as noted
> earlier. Order here is not significant.

michaelboulton:
I've added some pytests tests for the websocket handling now. They work on my machine with a few different versions of Python, but if they pass in CI then I think that's everything.

tests/test_client.py
```diff
@@ -147,3 +153,225 @@ def on_message(client, userdata, msg):
 
         packet_in = fake_broker.receive_packet(1)
         assert not packet_in  # Check connection is closed
+
+
+@pytest.mark.parametrize("proto_ver,proto_name", [
+    (client.MQTTv31, "MQIsdp"),
+    (client.MQTTv311, "MQTT"),
+])
+class TestInvalidWebsocketResponse(object):
```
Nzbuu: I wonder if it's better to put these websocket tests into a different file. This one is mostly about moving old tests to py.test.

tests/test_client.py
```diff
@@ -147,3 +153,225 @@ def on_message(client, userdata, msg):
...
+
+
+@pytest.mark.parametrize("proto_ver,proto_name", [
+    (client.MQTTv31, "MQIsdp"),
+    (client.MQTTv311, "MQTT"),
+])
+class TestBadWebsocketHeaders(object):
+    """ Testing for basic functionality in checking for headers """
+
+    def setup(self):
```
Nzbuu: py.test prefers test fixtures to `setup/teardown`.

tests/test_client.py
```diff
@@ -147,3 +153,225 @@ def on_message(client, userdata, msg):
 
         packet_in = fake_broker.receive_packet(1)
         assert not packet_in  # Check connection is closed
+
+
```
Nzbuu: How many of these tests could be replaced by tests on the `WebsocketWrapper` class directly, using a fake socket object? That should be much easier than creating a client and having it talk to a fake broker.

tests/test_client.py
```diff
@@ -1,9 +1,15 @@
 import os
 import sys
+import base64
+import re
 import inspect
+import hashlib
+from collections import OrderedDict
+from six.moves import socketserver
```
Nzbuu: Does `six` need to be added as a requirement?

tests/test_client.py
```diff
@@ -17,7 +23,7 @@
 import paho_test
 
 # Import test fixture
-from testsupport.broker import fake_broker
+from testsupport.broker import fake_broker, fake_websocket_broker
```
Nzbuu: Where is the fake_websocket_broker defined?

tests/test_client.py
michaelboulton: `six` is a dependency of pytest anyway. It won't be installed with the library, but it will be present when the tests are run.

tests/test_client.py
michaelboulton: Seeing as some of the tests are testing the behaviour of `ws_set_options` and the instantiation of the wrapper I thought it might as well create the whole client to make sure it passed the options through correctly. It could be changed but I think it might actually be more work

tests/test_client.py
michaelboulton: Makes sense, I just put them in there because the file already existed. I'll move it to test_websockets.py or something similar

tests/test_client.py
Nzbuu: OK

tests/test_client.py
Nzbuu: One problem with this approach is that the tests may fail even if it's nothing to do with WebsocketWrapper because you're relying on all of the client logic and the Python networking stack. These are integration tests rather than unit tests. The style of the other tests in Test_client is since they are a direct migration of the old tests, which were written in this way.

The way I see it, there are 3 groups of things that need to be tested:

1. `ws_set_options` stores the expected values in the client object and provides them to the WebsocketWrapper
2. WebsocketWrapper provides a socket-like interface to the client
3. WebsocketWrapper sends and receives the correct data with the socket it is wrapping

By using test doubles, you can avoid any network layer logic at all, with the possible exception of the first group. Also, most of these are entirely independent of MQTT.

tests/test_client.py
Nzbuu: Ah found it.

tests/testsupport/broker.py
```diff
@@ -53,7 +56,52 @@ def fake_broker():
     # print('Setup broker')
     broker = FakeBroker()
 
-    yield broker
+    try:
```
Nzbuu: I don't think that try-catch is necessary here. I think that py.test takes care of that.

michaelboulton: You're right, I didn't mean to commit that but it got caught up in the commit. For future reference from the [documentation](http://pytest.org/dev/yieldfixture.html) for anyone reading:

> With pytest fixture functions, the “after yield” part will always be invoked, independently from the exception status of the test function which uses the fixture.

tests/test_client.py
michaelboulton: Changed to use fixtures

tests/test_websockets.py
```diff
@@ -0,0 +1,242 @@
...
+    def test_unexpected_response(self, proto_ver, proto_name, fake_websocket_broker):
+        """ Server responds with a valid code, but it's not what the client expected """
+
+        mqttc = client.Client(
+            "test_unexpected_response",
+            protocol=proto_ver,
+            transport="websockets"
+            )
+
+        with fake_websocket_broker.serve("200 OK\n"):
```
PierreF: TCPServer don't expect a string. It cause the server to close the connection immediately.

tests/test_websockets.py
```diff
@@ -0,0 +1,242 @@
...
+            "test_correct_path",
+            protocol=proto_ver,
+            transport="websockets"
+            )
+
+        mqttc.ws_set_options(
+            path=mqtt_path,
+        )
+
+        def create_response_hash(decoded):
```
PierreF: This method is never used. I think it should be the "check_request" parameter of _get_callback_handler

tests/test_websockets.py
```diff
@@ -0,0 +1,242 @@
...
+            "test_correct_path",
+            protocol=proto_ver,
+            transport="websockets"
+            )
+
+        mqttc.ws_set_options(
+            headers=auth_headers,
+        )
+
+        def create_response_hash(decoded):
```
PierreF: This method is never used. I think it should be the "check_request" parameter of _get_callback_handler

michaelboulton:
@PierreF I've fixes those tests now. I also moved the 'integration' tests into another file so it's more obvious.

I had a go at writing some unit tests for ws_set_options but it's surprisingly complicated because it requires a large amount of mocking the socket library. I wrote one for the websocket wrapper but it also requires a lot of manual checking (and the behaviour changes between python 2 and 3).

PierreF:
Thanks for your contribution and thanks @Nzbuu for the help on review

PR #175, opened 2017-04-10
Merge v1.2.1 into fixes branch

Nzbuu:
Updates from master branch

PR #176, opened 2017-04-10
Merge v1.2.1 into develop branch

Nzbuu:
Brings develop branch up-to-date with fixes/master

All of the tests pass, so I can only assume that the changes have been merged correctly.

Note that this is based on my version of the fixes branch, which includes #175. So it might be worth rebasing this once #175 is merged so that the history is cleaner. **fixed**

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -500,6 +500,7 @@ class Client(object):
         self._last_retry_check = 0
         self._clean_session = clean_session
 
+        # [MQTT-3.1.3-4] Client Id must be UTF-8 encoded string.
         if client_id == "" or client_id is None:
             if protocol == MQTTv31:
                 self._client_id = base62(uuid.uuid4().int, padding=22)
@@ -1086,10 +1087,14 @@ class Client(object):
         Must be called before connect() to have any effect.
         Requires a broker that supports MQTT v3.1.
 
-        username: The username to authenticate with. Need have no relationship to the client id.
-        password: The password to authenticate with. Optional, set to None if not required.
+        username: The username to authenticate with. Need have no relationship to the client id. Must be unicode    
+            [MQTT-3.1.3-11].
+        password: The password to authenticate with. Optional, set to None if not required. If it is unicode, then it 
+            will be encoded as UTF-8.
         """
-        self._username = username
+
+        # [MQTT-3.1.3-11] User name must be UTF-8 encoded string
+        self._username = username.encode('utf-8')
         self._password = password
         if isinstance(self._password, unicode):
             self._password = self._password.encode('utf-8')
@@ -1766,8 +1771,7 @@ class Client(object):
                 else:
                     if len(byte) == 0:
                         return 1
-                    byte = struct.unpack("!B", byte)
-                    byte = byte[0]
+                    byte, = struct.unpack("!B", byte)
                     self._in_packet['remaining_count'].append(byte)
                     # Max 4 bytes length for remaining length as defined by protocol.
                     # Anything more likely means a broken/malicious client.
@@ -1796,8 +1800,8 @@ class Client(object):
             else:
                 if len(data) == 0:
                     return 1
-                self._in_packet['to_process'] = self._in_packet['to_process'] - len(data)
-                self._in_packet['packet'] = self._in_packet['packet'] + data
+                self._in_packet['to_process'] -= len(data)
+                self._in_packet['packet'] += data
 
         # All data for this packet is read.
         self._in_packet['pos'] = 0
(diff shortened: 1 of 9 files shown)


Nzbuu:
Heirarchy has been updated for #175. ip-validation is due to some of the included changes from fixes.

src/paho/mqtt/client.py
```diff
@@ -1734,7 +1764,10 @@ def _packet_read(self):
                     print(err)
                     return 1
                 else:
-                    byte, = struct.unpack("!B", byte)
+                    if len(byte) == 0:
+                        return 1
+                    byte = struct.unpack("!B", byte)
```
PierreF: Could you keep the unpacking with ```byte, = ...``` (added by #81) ?

src/paho/mqtt/client.py
```diff
@@ -1761,8 +1794,10 @@ def _packet_read(self):
                 print(err)
                 return 1
             else:
-                self._in_packet['to_process'] -= len(data)
-                self._in_packet['packet'] += data
+                if len(data) == 0:
+                    return 1
+                self._in_packet['to_process'] = self._in_packet['to_process'] - len(data)
```
PierreF: Could you keep the ```-=``` that is more readable ?

src/paho/mqtt/client.py
```diff
@@ -2043,19 +2077,22 @@ def _send_connect(self, keepalive, clean_session):
...
 
-        if self._username is not None:
-            remaining_length += 2 + len(self._username)
-            connect_flags |= 0x80
-            if self._password is not None:
-                connect_flags |= 0x40
-                remaining_length += 2 + len(self._password)
+        if self._username:
+            uusername = self._username.encode('utf-8')
+            remaining_length = remaining_length + 2 + len(uusername)
```
PierreF: Here also, ```+=``` is more readable.

src/paho/mqtt/client.py
Nzbuu: This is as it was in fixes.

src/paho/mqtt/client.py
Nzbuu: This is as it was in `fixes`.

src/paho/mqtt/client.py
Nzbuu: I can do this, but it looks like someone else replaced `+=` elsewhere.

src/paho/mqtt/client.py
```diff
@@ -1058,7 +1089,7 @@ def username_pw_set(self, username, password=None):
         username: The username to authenticate with. Need have no relationship to the client id.
         password: The password to authenticate with. Optional, set to None if not required.
         """
-        self._username = username.encode('utf-8')
```
Nzbuu: This is as it was in `fixes`.
But I agree with you: the handling of the unicode encoding/decoding is far from consistent.

Nzbuu: See https://github.com/eclipse/paho.mqtt.python/commit/ce27bf653a37911c595f64b2c843f7915645de8e

PierreF: Yes, because in fixes the .encoding was moved to _send_connect and in develop the encoding of password was added.
So it's the merge that caused encoding to be done in two place.

src/paho/mqtt/client.py
PierreF: Yes, because the new unpacking was added by #81 merged in develop and not fixes.
Therefor your merge is rolling back a part of #81.

src/paho/mqtt/client.py
PierreF: This also come from #81 merged only in develop.

src/paho/mqtt/client.py
PierreF: It's #81 that replaced ```value = value + offset``` by ```value += offset```.
#81 was only applied to develop not fixes.

src/paho/mqtt/client.py
Nzbuu: Should I keep the following from `master` (https://github.com/eclipse/paho.mqtt.python/commit/2539ea2d3ab1bf10e5e255f66dc61c09b996633b)?

```
                    if len(byte) == 0:
                        return 1
```

PierreF: Yes, that seems still relevant.

src/paho/mqtt/client.py
Nzbuu: I'm going to move the encoding here, because I prefer that too.

src/paho/mqtt/client.py
Nzbuu: Done

src/paho/mqtt/client.py
Nzbuu: Done

src/paho/mqtt/client.py
Nzbuu: Done

src/paho/mqtt/client.py
Nzbuu: Done

PR #179, opened 2017-04-10
Make auth dict handling a bit more lenient

mtreinish:
This commit changes the logic slightly on handling of the auth dict that
gets passed in to publish or subscribe methods. Previously they expected
None to always be what is passed in which was the default. However if an
empty dict was passed in this would fail with a KeyError. Additionally
the same was true if you pased a dict in without a username key. To make
this a bit more forgiving this changes the logic so if an object is
passed in that evaluates to False (like an empty dict) the object isn't
used. It also makes checking for the presence of 'username' and
'password' use the get() method so we don't raise blank Exceptions and
are a bit more explicit in usage.

This should be fully backwards compatible with any previous usage, but
hopefully make it a bit simpler for people playing with the lib for the
first time.

### DIFF
--- a/src/paho/mqtt/publish.py
+++ b/src/paho/mqtt/publish.py
@@ -123,13 +123,14 @@ def multiple(msgs, hostname="localhost", port=1883, client_id="", keepalive=60,
     client.on_publish = _on_publish
     client.on_connect = _on_connect
 
-    if auth is not None:
-        username = auth['username']
-        try:
-            password = auth['password']
-        except KeyError:
-            password = None
-        client.username_pw_set(username, password)
+    if auth:
+        username = auth.get('username')
+        if username:
+            password = auth.get('password')
+            client.username_pw_set(username, password)
+        else:
+            raise KeyError("The 'username' key was not found, this is "
+                           "required for auth")
 
     if will is not None:
         will_topic = will['topic']
--- a/src/paho/mqtt/subscribe.py
+++ b/src/paho/mqtt/subscribe.py
@@ -131,13 +131,14 @@ def callback(callback, topics, qos=0, userdata=None, hostname="localhost",
     client.on_message = _on_message_callback
     client.on_connect = _on_connect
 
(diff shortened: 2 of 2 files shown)


PierreF:
Thank you for your contribution.

I'm not really sure to understand the use case for supporting empty dictionary / dictionary without username. I don't see why you end up with a dictionary without username key and not with a None.

Either you are not providing credentials and auth should be None or you provide a username (optionally a password) and auth must be a dictionary with username key (and optionally a password). Somewhere in your code you will ends with:
```
if use_credentials():
    auth = {"username": source_for_username(), "password": source_for_password()}
else:
    auth = None
```

Also PR should be done on branch fixes or develop (fixes for bug fixes and develop for new feature).
PR #136 merged on develop branch simplified handling of auth parameter in publish/subscribe. It would be great to create your PR on the develop branch.
If you update the PR, please remember to sign the Eclipse CLA: https://www.eclipse.org/legal/ECA.php

mtreinish:
So it's not that you can't do it the current way just fine, it's more that current structure causes confusion for people not intimately familiar with the library. (because of the type change) A pretty typical case is with the way things currently are is if I write code like this:

```
def send_mqtt(conf, topic, msg):
    ...
    auth = None
    if conf.has_section('mqtt'):
        if conf.has_option('mqtt', 'username'):
            auth = {'username': conf.get('mqtt', 'username')}
            if conf.has_option('mqtt', 'password'):
                auth['password'] = conf.get('mqtt', 'password')
    ...
```

The first question I always see with that is why is auth = None in there instead of {}. People just expect that if it's taking a dict in the case where you set stuff it will handle it properly when you don't pass in the required fields.

Sure thing, I'll resubmit this on the develop branch, and I signed the Eclipse CLA so that should pass on the resubmission.

mtreinish:
Reopened as: https://github.com/eclipse/paho.mqtt.python/pull/184

PR #187, opened 2017-04-18
Use tox to run tests

Nzbuu:


Nzbuu:
The py27 test pass on my machine and on my fork on Travis: https://travis-ci.org/Nzbuu/paho.mqtt.python/builds/223156998

PierreF:
Yes, it's not tox that cause this timeout.

When I call python setup test, it produce timeout 4 run out of 10. So there is some issue with our pytest :/

Issue #186, opened 2017-04-18
print_topic in _handle_publish in develop branch


michaelboulton:
Just a minor thing - in the current develop branch the 'print_topic' variable in _handle_publish is assigned to but never used: 

https://github.com/eclipse/paho.mqtt.python/blob/fdb46b4e40c4c1a8d661ccb9a59b1975e6512db8/src/paho/mqtt/client.py#L2436-L2441

It used to print it out in the log message, but now it just prints the message topic.

master: https://github.com/eclipse/paho.mqtt.python/blob/master/src/paho/mqtt/client.py#L2456

develop: https://github.com/eclipse/paho.mqtt.python/blob/fdb46b4e40c4c1a8d661ccb9a59b1975e6512db8/src/paho/mqtt/client.py#L2455

PierreF:
Commit 6d9ce3a make use of print_topic to (re)fix #75 in develop.

PR #189, opened 2017-04-18
Tidy develop branch

Nzbuu:
* Update version number to 2.0.0.dev0
* Updates to changelog
* Ensure all classes are new-style
* Remove unused code (#186)
* Remove some unnecessary python 2/3 code
* Hard-code protocol name based on version number


PierreF:
I will wait before changing the version number. Using 2.0 or 1.3 or is not yet fixed.

For #186, this looks like a merge error. I mean that in develop an invalid UTF-8 topic will cause a crash instead when trying to print log message. It seems that commit 47348cfa was reverted.

Nzbuu:
I think that, given there are breaking changes, i.e. it may not work with old versions of Python, we should assume that it's going to be v2.x.

ChangeLog.txt
```diff
@@ -1,13 +1,14 @@
-v2.x - 2017-xx-xx
+v1.3.x - 2017-xx-xx
```
PierreF: The next version is not yet defined to be 1.3 or 2.0.
That why I would not yet update the version.

src/paho/mqtt/__init__.py
```diff
@@ -1,4 +1,10 @@
-__version__ = "1.2.2"
+__version__ = "1.3.0.dev0"
```
PierreF: The next version is not yet defined to be 1.3 or 2.0.
That why I would not yet update the version.

src/paho/mqtt/__init__.py
```diff
@@ -1,4 +1,10 @@
...
+
+VERSION_MAJOR = 1
+VERSION_MINOR = 3
+VERSION_REVISION = 0
+VERSION_NUMBER = (VERSION_MAJOR * 1000000 + VERSION_MINOR * 1000 + VERSION_REVISION)
```
PierreF: Those variable are not used by the library itself. Nor they are documented.
If they were used, moving them would be a breaking change, thus maybe an import in client.py could ease the migration.
That being said, I would said that because it's not used nor documented, we should just drop them completely.

src/paho/mqtt/client.py
```diff
@@ -2438,18 +2424,6 @@ def _handle_publish(self):
...
-        # representation of the topic for logging. When the user attempts to
-        # access message.topic in the callback, an exception will be raised.
-        if sys.version_info[0] >= 3:
-            try:
-                print_topic = topic.decode('utf-8')
-            except UnicodeDecodeError:
-                print_topic = "TOPIC WITH INVALID UTF-8: " + str(topic)
-        else:
-            print_topic = topic
-
```
PierreF: I've pushed change on develop that make use of print_topic to keep change from 47348cf

src/paho/mqtt/__init__.py
Nzbuu: I agree that they should probably be removed, since, if they are required, then they should be obtained from `__version__`.

However, defining them in the `client` module certainly feels like the wrong place, given it has sibling modules.

I thought that develop was now permitting minor breaking changes anyway.

src/paho/mqtt/__init__.py
Nzbuu: It makes me double-take every time I see tox output something like the following, when I'm really on the develop branch:

```
py36 inst-nodeps: C:\<snip>\paho_python\.tox\dist\paho-mqtt-1.2.2.zip
```

PierreF: Ok, got it. It's for installation of during development.
I feed it's another reason to drop VERSION_MAJOR/MINOR/REVISON/NUMBER since it's does not support "dev0" suffix.

swails: There are plenty of tools out there for doing versioning a bit more sensibly.  What I use for some of my projects is `versioneer` -- https://github.com/warner/python-versioneer.

Versions are parsed from git tags, and the version string includes the closest "tag" with the number of commits added since that tag, with the beginning of the commit hash that the install was cut from.  Pretty nice, I think.

Nzbuu: I remember seeing somewhere that it was bad practice to use the "__version__" attribute and importing it in `setup.py` since it importing the module executes the code in it. However, I can't remember where it was.

You don't need a __version__ attribute in order to determine the installed version of the package: https://github.com/RPi-Distro/python-gpiozero/issues/381

swails: `versioneer` also provides a convenient command class to inject into setup.py to handle that kind of thing.

Nzbuu: Update: The accepted answer here discusses why the `import` solution is bad: http://stackoverflow.com/questions/2058802/how-can-i-get-the-version-defined-in-setup-py-setuptools-in-my-package

ChangeLog.txt
Nzbuu: It can be updated once that decision has been made. This way it's just different from master/fixes

src/paho/mqtt/__init__.py
PierreF: The issue about dependency is only true if the module that contains version (and which is imported from setup.py) needs a dependency.
Here we import an "empty" __init__.py that contains only the version, which is fine.
That being said, it could indeed only be present in setup.py and being queried thought pkg_resource. This seems more Pythonic.

src/paho/mqtt/client.py
Nzbuu: OK. Unpicked this commit.

ralight:
Just as a comment, I have vague memories of a recommendation for Eclipse projects to use something like 1.2.99 as in-development version numbers but I haven't been able to find a reference for it so I might be mistaken.

src/paho/mqtt/__init__.py
Nzbuu: I agree. There isn't a problem for paho here (right now), but it is _technically_ bad practice.

I think there's also an argument that, if it is defined as an attribute, then it should be in the top-level module (`paho`) rather than in `paho.mqtt`.

Nzbuu:
@ralight , Python has specific support for development and pre-release versions: https://www.python.org/dev/peps/pep-0440/. So it seems best to use that by default, unless the top-level project prefers something different.

ralight:
@Nzbuu In that case I'd say stick with the Python .devX suffix as proposed.

PR #191, opened 2017-04-22
allow an arbitrary (not just /mqtt) path

levinanton:
Signed-off-by: Anton Levin <anton.levin@sap.com>
Change-Id: I2287f3020bf7379793d0bf009318a32c73e3d3f5

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -541,6 +541,7 @@ class Client(object):
         self._will_retain = False
         self._on_message_filtered = MQTTMatcher()
         self._host = ""
+        self._path = "/mqtt"
         self._port = 1883
         self._bind_address = ""
         self._in_callback = False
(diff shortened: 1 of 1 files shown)


rac01:
It would be great to have this pull request accepted! It's compliant to the MQTT standard that allows paths other than just /mqtt, it keeps all the existing functionality as a default and various parties already use it with the SAP IoT Starterkit https://github.com/SAP/iot-starterkit/tree/master/src/examples/python/mqtt-over-wss already - but have to apply an individual patch so far.

PierreF:
Thank you for your contribution. An alternative (PR #169) will also allow arbitrary websocket path in addition or arbitrary websocket header.
l'm closing this PR as the other one will provide more feature.

rac01:
Fully understood. Now it would be just great to see PR #169 proceed well. :-)

Issue #194, opened 2017-05-12
Incorrect handling of duplicate PUBLISH messages for qos 2


kelvinens:
If the client receives two or more PUBLISH messages with the same mid, it appends the message many times to the `_in_messages` list, **even on qos 2**.

This causes many calls to `_handle_on_message` with the same mid when receiving PUBREL.

PierreF:
PR #199 fix this issue.

PR #199, opened 2017-05-29
Fix duplicate callback on qos2 (#194)

kelvinens:
Verify if the message isn't already in the list before adding it (qos 2).
Messages with the same mid are considered equal.


PierreF:
Thank you for your contribution.
This looks good, could you please sign the CLA and sign-off your commit ?

kelvinens:
Fixed.

PR #205, opened 2017-06-26
Bind all callbacks as instance methods.

fquinto:
Similar idea from @yoch in #63 and some visual corrections

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -37,6 +37,7 @@ import uuid
 import base64
 import string
 import hashlib
+import types
 import logging
 
 try:
@@ -455,7 +456,7 @@ class Client(object):
     """
 
     def __init__(self, client_id="", clean_session=True, userdata=None,
-            protocol=MQTTv311, transport="tcp"):
+                 protocol=MQTTv311, transport="tcp"):
         """client_id is the unique client id string used when connecting to the
         broker. If client_id is zero length or None, then the behaviour is
         defined by which protocol version is in use. If using MQTT v3.1.1, then
@@ -722,7 +723,6 @@ class Client(object):
 
         Must be called before connect() and after either tls_set() or
         tls_set_context()."""
-
         if self._ssl_context is None:
             raise ValueError('Must configure SSL context before using tls_insecure_set.')
 
@@ -817,8 +817,8 @@ class Client(object):
         if keepalive < 0:
             raise ValueError('Keepalive must be >=0.')
         if bind_address != "" and bind_address is not None:
-            if (sys.version_info[0] == 2 and sys.version_info[1] < 7) or (
-                        sys.version_info[0] == 3 and sys.version_info[1] < 2):
+            if ((sys.version_info[0] == 2 and sys.version_info[1] < 7) or
+                    (sys.version_info[0] == 3 and sys.version_info[1] < 2)):
                 raise ValueError('bind_address requires Python 2.7 or 3.2.')
 
         self._host = host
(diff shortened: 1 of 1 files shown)


PierreF:
Tests are failing because this PR break callbacks (at least when not using subclass).
The following usage is what makes test fail:
```
def on_connect(mqtt_client, userdata, flags, rc):
    pass

mqtt_client.on_connect = on_connect
```

The callback is called with userdata, flag and rc but is missing mqtt_client first parameter.

swails:
Changes were made awhile ago to support subclassing `Client` that required a fair bit of weirdo Python magic to support more traditional subclassing behavior without breaking backwards compatibility.

What I realized doing that is that what is being attempted here is, quite simply, impossible to do while retaining backwards compatibility.

Since the API indicates that you should attach callbacks via assignment to *instances* (rather than the class definition itself), the callbacks *can't* be treated like an instance method (i.e., one in which the first argument is passed automatically as is typical for `self`).  If you can't treat these callbacks as instance methods, the only way to make the parent `Client` instance visible to callbacks in general is to leave them accepting a `client` argument that is redundant with `self` in the cases where those callbacks are instance methods.

While I like the feel of this API better, the churn it will impose on consumers of this client library that will be required to refactor to continue to use paho is significant and not worth the largely cosmetic API improvements IMO.

fquinto:
I don't have checked yet in deeply in order to say now that the compatibility backwards it's impossible. Maybe with "ugly and redundant" code can we get it? I understand the backward compatibility it's necessary but I need this subclass at the same time. Any specific proposal to support both needs?

swails:
> I understand the backward compatibility it's necessary but I need this subclass at the same time. Any specific proposal to support both needs?

Subclassing already works.  You just need to make sure that instance method callbacks have a redundant `client` argument which just gets ignored:

```python
class ClientSubclass(Client):

    def on_connect(self, _, userdata, flags, rc):
        # Your code goes here
```

Note you need paho 1.3.0 for this (overriding callbacks in subclasses did not work in earlier versions).

swails:
> I don't have checked yet in deeply in order to say now that the compatibility backwards it's impossible. Maybe with "ugly and redundant" code can we get it?

The only way you can maintain backwards compatibility is to change the places where the callbacks are called and use Python's introspection capabilities to look at how many arguments the method takes and whether it's an instance method or not, then based on the set of supported usage patterns use the appropriate call signature.

But this is beyond ugly -- it's confusing and rife with possible corner cases.  I can think of no other way of supporting both instance methods *and* non-instance methods in the same API.

fquinto:
Therefore... how to continue now? Implementation with _introspection capabilities in Python and counting how many arguments the method takes_ proposal, past the Travis test and accepting this _pull request_ or close this pull request? What do you think about that?

Clarification note: for me _ugly_ = look at how many arguments the method takes

swails:
My understanding is that the current implementation allows you to define instance methods for the callbacks in a subclass of `Client`, but requires you to accept a dummy `Client` instance redundant with `self` to support the original API that consisted of assigning callback functions to instances.

Basically what that means is that you need to change the call signature of your overriding callbacks from:

```python
class MyClient(Client):
    # This is "cleaner"
    def on_connect(self, userdata, flags, rc):
        <your code>
```

to:

```python
class MyClient(Client):
    # This isn't as "clean"
    def on_connect(self, _, userdata, flags, rc):
        <your code>
```

Unless I'm missing something and there's something else the changes proposed here allows that is otherwise harder to do, my suggestion is to close this PR and #63.

fquinto:
Thanks for your comments.

PR #206, opened 2017-07-07
Change iterable _userdata to deque

swails:
Some other cosmetic fixes.

Fixes #87 

Signed-off-by: Jason Swails <jason.swails@gmail.com>

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -144,8 +144,6 @@ MQTT_ERR_QUEUE_SIZE = 15
 
 sockpair_data = b"0"
 
-LOGGER = logging.getLogger(__name__)
-
 class WebsocketConnectionError(ValueError):
     pass
 
@@ -734,11 +732,14 @@ class Client(object):
             # If verify_mode is CERT_NONE then the host name will never be checked
             self._ssl_context.check_hostname = not value
 
-    def enable_logger(self, logger=LOGGER):
-        if self._logger is not None:
-            # Do not replace existing logger
-            return
-        self._logger = logger if logger is not None else LOGGER
+    def enable_logger(self, logger=None):
+        """ Enables a logger to send log messages to """
+        if logger is None:
+            if self._logger is not None:
+                # Do not replace existing logger
+                return
+            logger = logging.getLogger(__name__)
+        self._logger = logger
 
     def disable_logger(self):
         self._logger = None
@@ -1772,7 +1773,7 @@ class Client(object):
                     return MQTT_ERR_AGAIN
                 if err.errno == EAGAIN:
                     return MQTT_ERR_AGAIN
-                LOGGER.error('failed to receive on socket: %s', err)
+                self._easy_log(MQTT_LOG_ERR, 'failed to receive on socket: %s', err)
                 return 1
             else:
                 if len(command) == 0:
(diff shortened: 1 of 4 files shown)


src/paho/mqtt/client.py
```diff
@@ -353,10 +353,7 @@ def __ne__(self, other):
 
     @property
     def topic(self):
-        if sys.version_info[0] >= 3:
-            return self._topic.decode('utf-8')
-        else:
-            return self._topic
+        return self._topic.decode('utf-8')
```
swails: The version check is unnecessary here since `str` in Python 2 accepts `decode` just fine:

```python
Python 2.7.13 |Anaconda 2.5.0 (x86_64)| (default, Dec 20 2016, 23:05:08) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://anaconda.org
>>> u'some string'.decode('utf-8')
u'some string'
>>> 'some string'.decode('utf-8')
u'some string'
```

src/paho/mqtt/client.py
```diff
@@ -490,9 +487,14 @@ def __init__(self, client_id="", clean_session=True, userdata=None,
...
             raise ValueError('A client id must be provided if clean session is False.')
 
-        self._transport = transport
+        if transport.lower() not in ('websockets', 'tcp'):
+            raise ValueError('transport must be "websockets" or "tcp", not %s' % transport)
+        self._transport = transport.lower()
         self._protocol = protocol
-        self._userdata = userdata
+        if isinstance(userdata, collections.Iterable) and not isinstance(userdata, basestring):
+            self._userdata = collections.deque(userdata)
```
swails: docstrings indicate that `userdata` can be any type, but then it's treated inside `_do_publish` as an iterable with a `pop` method.  What should happen here?  Should `userdata` be enforced to be a deque and raise a `TypeError` if it's not an iterable that can be coerced into one?

Nzbuu: The client used in the publish module sets the userdata to be a queue. As you say, in the client module, it should be anything.

swails:
Thanks @Nzbuu!  I've made the appropriate changes.

I also changed the imports into relative imports with the appropriate `__future__` feature.  The upside of this is that `paho` can now be copied in its entirety as a subpackage, allowing users to bundle `paho` directly and eliminate it as a dependency *without* colliding with any other installed version of `paho` (or getting silently overwritten if a user then pip-installs paho separately).

There are also two instances where failed `isinstance` checks resulted in a `ValueError` being raised rather than the more sensible `TypeError`.  I changed these to the exception type that makes more sense, but it *does* technically change undocumented API features and potentially introduces some backwards-incompatible changes in a few corner cases.

Comments welcome.

swails:
Last change I just made was to replace uses of `print` with logging

swails:
Any other issues with this PR?

PierreF:
I don't like the logger change part of this PR. It introduce another logger:
* print() (ok this one no longer exists after this PR)
* existing on_log() method, that may using Python logging (en enable_logger is called)
* LOGGER introduced by this PR

I think it's better to use existing logger (_easy_log).

Could you remove the logger change from this PR (it's not releated to the PR, prefer to open multiple PR with one per subject it makes review easier). The other changes (PR related, pep8 fixes, import changes, and exception changes - ideally this would have been 4 PRs) looks good to me.

swails:
I made the logging change, now I'll split this into separate PRs.

PR #214, opened 2017-08-04
Make logging more consistent

swails:
This removes print statements and replaces them with appropriate logging calls.

logging


swails:
Comments?

PR #215, opened 2017-08-04
Turn list of messages to publish into a deque

swails:
Fixes #87 (replaces #206)

stack


swails:
Note this also turns the imports into relative imports, which allows `paho` to be embedded as a subpackage in other projects.

I'd have put this in a separate PR, but it would have conflicted with this one and I didn't feel like dealing with the headache.

swails:
Comments?

PR #218, opened 2017-08-17
Fix reconnect_delay_set

philipbl:
Sorry if I am missing something in the code or if I don't have a good understanding of MQTT. Previously `self._reconnect_max_delay` was set to `min_delay` when `reconnect_delay_set` was called. Now `self._reconnect_max_delay` is set to `max_delay`.


PierreF:
Yes, it was a bad copy/paste. Thanks for catching this.

PR #219, opened 2017-08-18
Add clean_session to callback and simple subscribers

philipbl:
The `clean_session` option for a Client was not exposed in `subscribe.callback` and `subscribe.simple`. This adds that option for both of these functions.


PierreF:
This PR create a breaking change: if someone called those method using positional parameter, the position have changed.
We don't want to introduce breaking change in fixes branch.

So we can either go for a breaking change and apply this PR on develop, or we can make this change a non-breaking change by adding this parameter as last one.
I would opt for the second option.

swails:
I agree. There isn't a strong justification for a breaking change, so i would opt for just putting the new parameter last.

philipbl:
Yeah, sorry about that. I will add it to the end of the parameter list.

PR #227, opened 2017-09-08
Make pytest-runner a conditional requirement

yegorich:
Check for pytest-runner only if setup.py was invoked with 'test'
argument.

Signed-off-by: Yegor Yefremov <yegorslists@googlemail.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,8 @@ with open('README.rst', 'rb') as readme_file:
 
 requirements = []
 test_requirements = ['pytest', 'pylama']
-setup_requirements = ['pytest-runner']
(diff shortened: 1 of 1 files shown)


PierreF:
Thank for your PR, but a same PR is already merge on fixes branch (#207). It is not yet released.

For future reference, PR should be done either on branch fixes or develop (not on master):
* fixes is intended for small backward compatible change that will make the next "fixes" release (e.g. 1.3.1, 1.3.2, ...)
* develop is indented for heavier change or backward incompatible change. It will goes in 1.4 or 2.0 release.

PR #230, opened 2017-09-11
Add __enter__ and __exit__ to use client in a context manager

sieben:


sieben:
I've signed the eclipse validation. Can you relaunch the check?

PierreF:
Some documentation would be appreciated.

Also currently the context manager does not:
* start the loop when entering the context
* stop the loop when exiting the context
* wait anything (calling disconnect is asynchronous. IIRC stopping the loop does wait for disconnect, but not necessary for all messages to be sent)

Finally, doesn't the subscribe/publish helper already provide a better way to support use case that need a context manager ? I mean if the use case is:
* connect to a MQTT broker
* send some messages
* disconnect from the broker
Then the helper seems better fitted than adding a context manager. Have you some use case that are more easily solved with a context manager ?

PierreF:
Closing as this PR seems staled. Feel free to reopen if you are still working on it.

Issue #147, opened 2017-02-01
Improve documentation for external event loop


ezramorris:
Related: https://github.com/eclipse/paho.mqtt.python/issues/72

I think the documentation could do with some clarification/small additions for external event loop usage.

Whilst incoming MQTT data is quite self-explanatory (set the event loop to watch for read availability on the socket(), and then run loop_read()), the writing of data is less clear:
- Is there any need to watch the socket() for write availability, and call loop_write() as a result? (I'm guessing not, but the docstring for loop_write() seems to suggest otherwise.)
- What's the correct way to wait on new outgoing data? The two examples contributed in the above issue ([1](https://github.com/eclipse/paho.mqtt.python/issues/72#issuecomment-224860637) and [2](https://github.com/eclipse/paho.mqtt.python/issues/72#issuecomment-263196153)) take different approaches: the first just polls want_write() to determine whether to call loop_write(), and the second watches the internal _sockpairR socket. The second option seems to be more suited to external event loops*.

I would be willing to draft some documentation, with an example using [asyncio](https://docs.python.org/3.4/library/asyncio.html), if these points could be clarified.

*On a side note, if this is the preferred option for waiting for new outgoing packets, should there be a public interface for this?

downwith:
There's more than one non-public member in the second example.  _state was being used too.  I was thinking that using asyncio to manage this and other sockets would be piece of cake as well.  Now I'm second guessing myself.

joernheissler:
In my opinion, the existing documentation on external event loops is sufficient. That is, if you know in general how the select system call work. In a nutshell: You can ask the operating system if there is anything to read from a file descriptor or if there's enough buffer space for writing something to a file descriptor. And select will block until the answer is "yes" to at least one of the questions (or a timeout is hit).

> Is there any need to watch the socket() for write availability, and call loop_write() as a result? (I'm guessing not, but the docstring for loop_write() seems to suggest otherwise.)

Yes, there is. There are 4 possible situations here: You want to write something, or you don't. There's space for writing or there isn't.

  * There's space and you want to write: You should call loop_write.
  * There's space and you don't want to write: loop_write will consume CPU time for nothing. Don't call it. And if there's nothing to do at all, you should be blocking or select until something interesting happens.
  * There's no space but you want to write: Don't call loop_write. Wait (i.e. block on select) until there's space or you'll waste CPU time.
  * There's no space and you don't want to write: Same.

> What's the correct way to wait on new outgoing data?

[...]

Issue #72, opened 2016-06-08
Add External Event Loop Example


zpfvo:
Hi,
i'd like to use this library together with the twisted reactor. Since i'm new to both paho.mqtt and twisted  its hard for me to figure this one out. It would be nice to have an example of the external event loop usage (mustn't be twisted).

ralight:
That sounds like a reasonable request, is there anybody out there that has an example they could share?

zpfvo:
I got it running, but i don't think its very nice. I don't use Deferreds and I don't use select on the socket.
(not a complete example)
```python
from twisted.internet import reactor, task 

def is_connected(mqttc):
    if mqttc.socket() is None:
        logging.error("Connection to mqtt broker lost. Stopping")
        reactor.stop()

def want_write(mqttc):
    if mqttc.want_write():
        mqttc.loop_write()

# MQTT STUFF
mqttc = mqtt.Client(transport="websockets")
mqttc.on_message = on_message
mqttc.on_connect = on_connect
mqttc.on_publish = on_publish
mqttc.on_subscribe = on_subscribe
# Uncomment to enable debug messages
# mqttc.on_log = on_log
mqttc.connect("localhost", 1883, 60)
mqttc.subscribe("events/#", 0)
mqttc.message_callback_add('events/comp', on_component_event_message)

# mqtt event loop functions
task.LoopingCall(mqttc.loop_read).start(0.1)
task.LoopingCall(want_write, mqttc).start(0.1)
task.LoopingCall(mqttc.loop_misc).start(5)
task.LoopingCall(is_connected, mqttc).start(1)

reactor.run()
```

oxbambooxo:
i'am adaptive it for gevent
(not a complete example)
```python
import paho.mqtt.client as mqtt
import gevent

def handle_write(mqttc):
    mqttc._sockpairR.recv(1)
    status = mqttc.loop_write()
    if status != mqtt.MQTT_ERR_SUCCESS:
        print('error', mqtt.error_string(status))
        mqttc._state = mqtt.mqtt_cs_disconnecting


def handle_read(mqttc):
    status = mqttc.loop_read()
    if status != mqtt.MQTT_ERR_SUCCESS:
        print('error', mqtt.error_string(status))
        mqttc._state = mqtt.mqtt_cs_disconnecting


def loop_forever(mqttc, idle=3):
    while mqttc._state != mqtt.mqtt_cs_disconnecting:
        status = mqttc.loop_misc()
[...]

Issue #204, opened 2017-06-25
Support for Ubuntu 14.04 users cut off


NickWaterton:
The recent breaking change to not support versions of python lower than 2.7.9 has left all Ubuntu 14.04 users in the cold.

We are at python 2.7.6, and trying to upgrade would break all our dependencies. I have to stay with V1.2.3 in order to continue to use this library.

I am using TLS V1.1, and I understand that python 2.7.9 is required for TLS V1.2, but I don't understand why V1.3 is not backward compatible with the thousands of Ubuntu 14.04 users out there.

I should be able to use TLS 1.1, but now I can't. Backwards compatibility with existing users should be a priority.

PierreF:
Hi,

support for new features like SNI or more powerful SSL context changed the code that handle TLS setup. New code rely on SSL context. Supporting both old and new method create a more complicated code and make it harder to maintain.

That's why we dropped support for older Python version.

That being said, Ubuntu 14.04 user has the option to go with Python 3.4 which is supported. If your application can't yet run on Python 3, backports.ssl package may be a solution.

NickWaterton:
So you believe that instead of a "harder to maintain" source code library, everyone else should simply:

- Change their operating system
- Change their programming language
- Rewrite their whole program to accommodate your changes.

Whereas I think that you should maintain backwards compatibility.

ralight:
Hi Nick,

You can still get the previous version. To install that specific version with pip, do:

    pip install paho-mqtt==1.2.3

swails:
@NickWaterton - There are many things you can do to use the `paho-mqtt` library on Ubuntu 14.04:

1. As @ralight suggested, use `pip install paho-mqtt==1.2.3` to get a version compatible with your system-level Python
1. Use `virtualenv` to install a Python 2.7.9+
1. Install Python 3.4 using `apt-get` (as suggested by @PierreF)
1. Install the latest version of Python 2.7 from a commercially-backed distribution like miniconda (this is very easy)

There are more than enough (very easy) workarounds that do not require you to change your OS. 
[...]

Issue #188, opened 2017-04-18
Incomplete handling of some socket errors


jamesmyatt:
There are a couple of places where the error number associated with a `socket.error` is checked against `(errno.EINPROGRESS, errno.EWOULDBLOCK, EAGAIN)` and all other exceptions are re-raised. However, when the error number is in this list, there is no action, which is probably wrong.

For example:

* https://github.com/eclipse/paho.mqtt.python/blob/develop/src/paho/mqtt/client.py#L859
* https://github.com/eclipse/paho.mqtt.python/blob/develop/src/paho/mqtt/client.py#L258

PierreF:
The one in reconnect seems fine for me.
If only means that reconnect in fact didn't reconnected, and self._sock stay None, which only cause later call to loop() to re-try connection.

The one in _socketpair_compat could be problematic. I agree that if such error occur, I think later code will fail (probably hang). That being said, I not sure those error could occur since it's using localhost, which should not fail.

engineeringentropy:
Looks like this is still an issue in 1.4.0:
ERROR:__main__:Unknown exception while waiting for conclusion of application: local variable 'sock' referenced before assignment
Traceback (most recent call last):
  File "databaser.py", line 81, in <module>
    databaser = MqttDatabaser(caFile=args.ca, keyFile=args.key, certFile=args.cert, queue=dataQueue)
  File "/mnt/c/Users/.../mqtt/databasermqtt.py", line 53, in __init__
    mqtt.connect("XXXX", 8883, 10)
  File "/mnt/c/Users/..../venv/lib/python3.6/site-packages/paho/mqtt/client.py", line 839, in connect
    return self.reconnect()
  File "/mnt/c/Users/..../venv/lib/python3.6/site-packages/paho/mqtt/client.py", line 974, in reconnect
    sock,
UnboundLocalError: local variable 'sock' referenced before assignment.

The above was obtained by dropping a VPN connection before running the application. 
sock = socket.create_connection((self._host, self._port), source_address=(self._bind_address, 0)) throws EAGAIN, which then falls through to 
sock = self._ssl_context.wrap_socket(
                    sock,
a few lines later.

I'm not sure what the correct behaviour should be.

engineeringentropy:
For anyone else getting this error with certificates where the server either disappears or isn't found:
The exception handler at client.py:963 can be changed to reraise the socket.error (remove the if-clause), and then let the background loop catch the error and run with it.

[...]

PR #235, opened 2017-10-08
Improve support for external event loops.

joernheissler:
#72,  #147, https://dev.eclipse.org/mhonarc/lists/paho-dev/msg03997.html and https://github.com/mossblaser/aiomqtt demonstrate that the current support for external event loops should be improved.

https://dev.eclipse.org/mhonarc/lists/paho-dev/msg03999.html suggests to add new callbacks to track the state of the library. That's what my change implements.

### DIFF
--- a/examples/loop_select.py
+++ b/examples/loop_select.py
@@ -1,0 +1,89 @@
+#!/usr/bin/env python3
+
+import socket
+import uuid
+import paho.mqtt.client as mqtt
+from select import select
+from time import time
+
+client_id = 'paho-mqtt-python/issue72/' + str(uuid.uuid4())
+topic = client_id
+print("Using client_id / topic: " + client_id)
+
+
+class SelectMqttExample:
+    def __init__(self):
+        pass
+
+    def on_connect(self, client, userdata, flags, rc):
+        print("Subscribing")
+        client.subscribe(topic)
+
+    def on_message(self, client, userdata, msg):
+        if self.state not in {1, 3, 5}:
+            print("Got unexpected message: {}".format(msg.decode()))
+            return
+
+        print("Got message with len {}".format(len(msg.payload)))
+        self.state += 1
+        self.t = time()
+
+    def on_disconnect(self, client, userdata, rc):
+        self.disconnected = True, rc
+
+    def do_select(self):
+        sock = self.client.socket()
+        if not sock:
+            raise Exception("Socket is gone")
+
+        print("Selecting for reading" + (" and writing" if self.client.want_write() else ""))
+        r, w, e = select(
+            [sock],
+            [sock] if self.client.want_write() else [],
+            [],
+            1
+        )
+
+        if sock in r:
+            print("Socket is readable, calling loop_read")
+            self.client.loop_read()
+
+        if sock in w:
+            print("Socket is writable, calling loop_write")
+            self.client.loop_write()
+
+        self.client.loop_misc()
+
+    def main(self):
+        self.disconnected = (False, None)
+        self.t = time()
+        self.state = 0
+
+        self.client = mqtt.Client(client_id=client_id)
+        self.client.on_connect = self.on_connect
+        self.client.on_message = self.on_message
+        self.client.on_disconnect = self.on_disconnect
+
+        self.client.connect('iot.eclipse.org', 1883, 60)
+        print("Socket opened")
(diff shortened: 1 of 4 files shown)


TD22057:
@joernheissler Just wanted to say thanks for doing this.  This is going to be very useful. :thumbsup:

src/paho/mqtt/client.py
```diff
@@ -567,16 +585,60 @@ def __init__(self, client_id="", clean_session=True, userdata=None,
...
+            if self._ssl and err.errno == ssl.SSL_ERROR_WANT_READ:
+                raise WouldBlockError()
+            if self._ssl and err.errno == ssl.SSL_ERROR_WANT_WRITE:
+                self._call_socket_register_write()
+                raise WouldBlockError()
+            if err.errno == EAGAIN:
+                raise WouldBlockError()
+            raise
+
+    def _sock_send(self, buf):
```
Nzbuu: I think that this addresses issue #188 as well.

examples/loop_select.py
```diff
@@ -0,0 +1,87 @@
...
+class SelectMqttExample:
+    def __init__(self):
+        pass
+
+    def on_connect(self, client, userdata, flags, rc):
+        print("Subscribing")
+        client.subscribe(topic)
+
+    def on_message(self, client, userdata, msg):
+        if self.state not in {1,3,5}:
```
Nzbuu: I would prefer to see the positive case first, so that the else clause doesn't mean "not not in {...}".

lyssdod:
Any chance that would be merged and included in next release?

examples/loop_select.py
joernheissler: I removed the "else". I prefer to have a program flow where the "good" case is always on the top indention level and all the bad cases just exit the function as soon as possible. "if this is bad, abort. If that is bad, abort. What's remaining must be good."

src/paho/mqtt/client.py
joernheissler: I don't really understand #188; I don't see the relation to my change. Maybe I looked at the wrong code lines?

joernheissler:
I made some minor improvements to the documentation, and loop_misc in the asyncio example is now started from a callback.

PierreF:
Thanks for this PR, it add good support for asyncio  to paho-mqtt.

While reviewing it, I'm wondering if our multiple loops support is not too complicated. My view is that

* loop_start: is the easiest, user don't care about details of how it works
* loop_forever: is easy also, user want to wait for completion
* custom loop based on loop_read/loop_write/loop_misc: complex solution, not very tested (it's with this PR that correct support for asyncio arrive). The use-case (in my view) is to use asyncio with lots of connection.

My question is then: should we keep all those loops ? What if the default loop is an
asyncio loop ?
* This would render custom loop useless
* loop_forever could be a call to asyncio.run_until_complete
* loop_start could be a thread that call loop_forever

Did you see any issue with going that way ? Any use case that require a custom base loop that a default asyncio won't match ?

joernheissler:
> My question is then: should we keep all those loops ? What if the default loop is an asyncio loop?

Asyncio was only added in py 3.4. Unless you want to anger your users (#204) you shouldn't depend on asyncio. Personally I don't care about py < 3.5, but I suspect others do.

The socket() and want_write() methods could be removed in my opinion. Instead, a select style loop can use the new callbacks and keep two lists of file descriptors that need watching (reading+writing). If a callback is called, those lists are modified.

What you could do too is move the existing loops from the Client class to an extra module and perhaps keep a stub method with the old signature so existing code won't break. Those loops would use the new callbacks.

PierreF:
I though that asyncio existed on PyPI for older Python... but that not the case. It could be possible to fallback on select based loop that being said.
I will merge this PR and see later if the existing loop could make use of the new callbacks.
Thanks for your work and patience !

Nzbuu:
The develop branch only supports Python 2.7 or 3.4+, and you can use https://github.com/haypo/trollius for Python 2.7 for asyncio.

It certainly feels like supporting the standard Python mechanism for asynchronous I/O would be a significant improvement.

Nzbuu:
Actually, looks like trollius is abandoned and probably wouldn't be very helpful anyway since it would lack the keyword support.

src/paho/mqtt/client.py
liusong1111: anonying.

Issue #247, opened 2017-11-15
Pub/Sub Samples with TLS, username, password etc. 


jonquark:
I'm thinking of writing (and contributing) samples that set up certs for TLS, support usernames/passwords, clientids (configured by parsing command line arguments to the sample).

Would such samples be welcome and if so, do you think I should extend client_sub.py or keep that as it is and add a client_sub_opts.py (which would take arguments for certs file, host, post, username)?

PierreF:
Yes, go ahead, we don't have yet example with TLS.

About creating a new client_sub_opts.py (maybe client_sub_tls ?) or updating client_sub.py, I would said it depend on complexity added.
I think it good to have a "simple" client_sub.py that is easy to understand. If adding option keep it simple enough it's probably better to keep number of example small and update client_sub.py

PierreF:
PR #252 merged

PR #251, opened 2017-12-01
Add samples with username & password & TLS  eclipse/paho.mqtt.python#247

jonquark:
Add samples as mentioned in:
https://github.com/eclipse/paho.mqtt.python/issues/247


jonquark:
Will redo.

PR #253, opened 2017-12-04
client: fix keepalive rule

iblis17:
Says, we have a client just keep pushing something the server,
it doesn't intent to receive anything.
So I alter the checking rule form `or` to `and`

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -1945,7 +1945,7 @@ class Client(object):
             last_msg_out = self._last_msg_out
             last_msg_in = self._last_msg_in
 
-        if self._sock is not None and (now - last_msg_out >= self._keepalive or now - last_msg_in >= self._keepalive):
+        if self._sock is not None and (now - last_msg_out >= self._keepalive and now - last_msg_in >= self._keepalive):
             if self._state == mqtt_cs_connected and self._ping_t == 0:
                 self._send_pingreq()
                 with self._msgtime_mutex:


PierreF:
I think the logic is intended to be an or.
For the standard client are only required to send pingreq between outbound package (e.g. it does not need to watch for last_msg_in at all).
But watching for last_msg_in will allow to detect a broken connection. You may be pushing to a blackhole, therefor the keepalive on incoming data will detect it (yes TCP would also do it... but slower than the default keepalive of 60 seconds).

Do you have any issue with the current keep-alive ?

Nzbuu:
The keep alive process is described in section of 3.1.2.10 the MQTT 3.1.1 spec. It says that the keep alive time "is the maximum time interval that is permitted to elapse between the point at which the Client finishes transmitting one Control Packet and the point it starts sending the next".

Hence, I think that it only refers to the last time that the client _sent_ a control packet of any kind and has no bearing on the last time a client _received_ a control packet.

So, in the extreme, if a client is only subscribed to topics at QoS 0 and doesn't publish anything, then it will only _send_ PINGREQ packets and this will be every keep alive interval, regardless of the amount of traffic from those subscriptions.

iblis17:
> Do you have any issue with the current keep-alive ?

here is my sample code, ran with python 3.6

```python

import time
import datetime

    print("Connected with result code "+str(rc))

client = mqtt.Client()

client.connect("localhost", 1883, keepalive=1)

client.loop_start()

while 1:
    d = str(datetime.datetime.now())
    p = client.publish("test", d, qos=0)
```

and I monitor it with
```
$ mosquitto_sub -t test
```

MQTT server is `mosquitto version 1.4.14` on FreeBSD 
```
$ uname -a
FreeBSD abeing 12.0-CURRENT FreeBSD 12.0-CURRENT #0 r323335: Sun Sep 17 00:56:35 CST 2017     root@abeing:/usr/obj/usr/src/sys/GENERIC  amd64
```

It just quite easy ran into reconnecting:
```
└─[iblis@abeing]% python3 test.py                             
Connected with result code 0  
[Errno 32] Broken pipe        
Connected with result code 0  
[Errno 32] Broken pipe        
[Errno 32] Broken pipe        
Connected with result code 0  
[Errno 32] Broken pipe        
[Errno 32] Broken pipe        
Connected with result code 0  
[Errno 32] Broken pipe        
Connected with result code 0
```

And my observation is: lower keeplive (e.g.  set it to `keepalive=1`) will make it reconnect in higher frequency.

(my patch seems doesn't help for this issue now, I made this patch long time ago....maybe I recall it wrongly)

iblis17:
well, if I set `keepalive=0` in the script I posted, it's happy and without any reconnection.
I believe there is something strange  in `_check_keepalive`.

PierreF:
Your sample will saturate CPU (and probably memory) because you are publishing as fast as possible.
You should add a time.sleep() in your loop.

You should also avoid so small keepalive. Mosquitto won't support such small keepalive, mostly because it is using seconds (as integer) to track time. Therefor 1 second is the smallest interval of time known by mosquitto (which means that even if the last message received 1 microsecond ago, but on the previous second mosquitto will think the message is 1 full second old).

Are you able to reproduce this issue without saturating resources and without such small keepalive ? I don't have any issue with your script after adding a time.sleep(0.01) and a keepalive of 5 seconds.

iblis17:
@PierreF Thanks for your patient reply. It's really informative.
We have some needs for a long run application.
My labmate always bugs me that our application cannot keep connection overnight.
I will do more tests with larger `keepalive` value.

so... I'm going to close this PR first.
If we have new find, I will post here.

Jyneda:
I set keepalive = 60, nothing changed.  My mqtt-client disconnected with server after about 90 minutes, but the mqtt-client never detects the disconnect state; meanwhile, the memory usage is constant increased.

PR #258, opened 2018-01-09
Allow resetting the username back to None

bapowell:
In client.py, update username_pw_set() to allow resetting the username back to None.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -1199,7 +1199,7 @@ class Client(object):
         """
 
         # [MQTT-3.1.3-11] User name must be UTF-8 encoded string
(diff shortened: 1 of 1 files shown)


PierreF:
LGTM.
Travis errors are unrelated (and transient I think).
You will need to pass ip-validation before I can merge this:
* You need to sign the Eclipse contributor agreement (https://www.eclipse.org/legal/ECA.php on the left ECA > Sign)
* Your commit need to be signed-off (git commit -s option).

I would be also great to update doc/docstring to state that username could be reset this way (not mandatory, I could do this later).

bapowell:
Will issue another pull request, with, using proper sign-off.

PR #259, opened 2018-01-10
Allow resetting the username back to None

bapowell:
In client.py, update username_pw_set() to allow resetting the username back to None.

Signed-off-by: Brad Powell <bapowell@sbcglobal.net>


PierreF:
Thanks. I don't known why Travis failed, It works on local.

Issue #248, opened 2017-11-20
Websocket over TLS can't send message larger than 16384 bytes


calebhalvy:
I have tls set and am connecting using websockets to the broker, but can't send messages that are very large, i.e. > 16384(2^14) bytes. When I don't set connection as secure, the messages send. I get an errno 11 inside _buffered_read in client.py at 
` if len(data) < wanted_bytes:
        raise socket.error(errno.EAGAIN, 0)`

PierreF:
I can't reproduce this issue. I'm able to send and receive message up to 1 Mb over TLS websocket.
Could you provide more information ? The full traceback will help.
More details on what you are doing may also help (I'm not sure to see why *sending* message cause an error on a *read* function), ideally a small example that exhibit this issue.

larshesel:
Hi,

Not 100% sure this is related, but we had a similar issue, but the problem was when the websocket payload encoding was between 32768  and 65535. This seems to be an issue here:

 https://github.com/eclipse/paho.mqtt.python/blob/master/src/paho/mqtt/client.py#L2805-L2814

Particularly the `elif length < 32768` case as the spec states: `Note that in all cases, the minimal number of bytes MUST be used to encode the length` and as we can encode values up to 16 bits using the 7 bits (val 126) + 16 bit encoding we should use that instead of the 64 bit encoding to conform to the spec.

I'll create a PR for this case in a bit.

larshesel:
yeah, I could imagine most websocket server implementations would be lenient about this particular part of the spec. I experienced this problem using the websocket implementation in the Cowboy web-server which we use in VerneMQ.

ralight:
I've not been able to reproduce the problem with the fix from #267, so I'm closing this. If it's still a problem, please reopen.

PR #266, opened 2018-01-29
Fix websocket payload length encoding

larshesel:
The spec says that the payload encoding MUST be the smallest
possible. As we can encode anything in [127;65536] in 16 bits (the 7 +
16 bit case in the spec, page 29), we should do this instead of
encoding [32768;max] using 64 bits.

I added a comment in the maybe related issue: https://github.com/eclipse/paho.mqtt.python/issues/248#issuecomment-361207407


larshesel:
Replaced this PR with #267 (sorry, didn't read the contributor guidelines well enough the first time around)

PR #267, opened 2018-01-29
Fix websocket payload length encoding

larshesel:
The spec says that the payload encoding MUST be the smallest
possible. As we can encode anything in [127;65536] in 16 bits (the 7 +
16 bit case in the spec, page 29), we should do this instead of
encoding [32768;max] using 64 bits.

Signed-off-by: Lars Hesel Christensen <lars@larshesel.dk>

Replaces #266


PierreF:
Thanks, good catch

PR #270, opened 2018-02-03
Add long type to type check

nbougues:
It seems to me that "long" has been forgotten here.


PierreF:
Yes, I think long should be also present here.
Could you fix few issues before that PR could be merged:

* As you may have seen, this break Python3, as Python3 don't have long type. You could use the same compatibility trick done for basestring and unicode in beginning of client.py
* You will need to sign Eclipse contributor agreement (https://www.eclipse.org/legal/ECA.php) and have a signed-off commit (use commit -s)
* This PR need to be done on the fixes branch

ralight:
Thanks to Python 2.7 no longer being important this PR isn't really needed any more. Thank you for reporting it anyway though.

Issue #276, opened 2018-03-08
MQTT client fails to reconnect properly if connection is lost while heavily publishing on QoS 2


amrbekhit:
Hello,

I've got some code which is frequently publishing messages using QoS 2. I noticed that if the connection is lost during a transmission, the library will attempt to reconnect later on, but then immediately disconnect. This gets seen on the server side as follows (using mosquitto-1.4.15):
```
Mar 08 07:41:42 localhost mosquitto[10327]: New connection from xx.xx.xx.xx on port 1883.
Mar 08 07:41:42 localhost mosquitto[10327]: New client connected from xx.xx.xx.xx as cf98f1b4-293d-4514-ab6e-9731b1462ea5 (c1, k60, u'username').
Mar 08 07:41:42 localhost mosquitto[10327]: Sending CONNACK to cf98f1b4-293d-4514-ab6e-9731b1462ea5 (0, 0)
Mar 08 07:41:42 localhost mosquitto[10327]: Socket error on client cf98f1b4-293d-4514-ab6e-9731b1462ea5, disconnecting.
```
I can replicate the problem very regularly using the below script:

```
import paho.mqtt.client as mqtt
import logging
import threading
import random
import string
import time


class MQTTStressTest:
    USERNAME = "username"
    PASSWORD = "password"
    SERVER = "mqttserver"

    def __init__(self):
        logging.getLogger().setLevel(logging.INFO)

        self.mqtt_client = mqtt.Client()
        self.mqtt_client.username_pw_set(self.USERNAME, self.PASSWORD)
        self.mqtt_client.on_connect = self.__mqtt_on_connect
        self.mqtt_client.on_disconnect = self.__mqtt_on_disconnect
        self.mqtt_client.on_message = self.__mqtt_on_message
        self.mqtt_client.on_publish = self.__mqtt_on_publish
        self.mqtt_client.parent = self

        self.is_connected = threading.Event()

    @staticmethod
    def __mqtt_on_publish(client, userdata, mid):
        logging.info("Message {} published.".format(mid))

    @staticmethod
    def __mqtt_on_disconnect(client, userdata, rc):
        logging.warning("============================MQTT disconnected.")

    @staticmethod
    def __mqtt_on_connect(client, userdata, flags, rc):
        self = client.parent
        if rc == mqtt.CONNACK_ACCEPTED:
            logging.info("===========================MQTT connected.")
            self.is_connected.set()
        else:
[...]

PR #281, opened 2018-03-21
Completely republish QoS 2 messages when reconnecting clean-session clients

amrbekhit:
Modified _messages_reconnect_reset_out() to completely republish QoS 2 messages when a clean-session client reconnects.

This eliminates issue #276 whereby clean-session clients would previously attempt to resume transmission of QoS 2 messages where they left off. However, because the server does not save state for such clients, the messages are unexpected and causes the server to close the connection (observed using mosquitto-1.4.15). This then continuously prevents the client from reconnecting.

loop_misc() has also been modified to only call _message_retry_check() if the client is connected. This prevents the client from immediately retrying transmission after sending a CONNECT message but before receiving CONNACK from the server. The out-of-order messages again causes the server to close the connection, preventing the client from reconnecting.

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -1,10 +1,12 @@
-v1.3.1 - 20??-??-??
+v1.3.1 - 2017-10-09
 ===================
 
 - Fix reconnect_delay_set which ignored the max_delay. Closes #218.
 - Fix crash when connection is lost while trying to send message. Closes #208.
 - Fix issue with unicode topic when some character were multi-bytes UTF-8.
 - Fix issue with empty Client ID with broker that don't support them. Closes #209.
+- Fix issue with tls_set that did not allowed cert_reqs=ssl.CERT_NONE.
+  Closes #123.
 - Relax requirement on pytest-runner, it's now only required for tests. Closes
   #207, #227.
 
--- a/src/paho/mqtt/__init__.py
+++ b/src/paho/mqtt/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "1.3.0.dev0"
+__version__ = "1.3.1"
 
 
 class MQTTException(Exception):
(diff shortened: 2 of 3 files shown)


vrst37:
You will need to sign Eclipse contributor agreement (https://www.eclipse.org/legal/ECA.php) and have a signed-off commit (use commit -s) before this commit can be accepted. 

Also, these changes have failed the test on Travis. I think these would have to be fixed before the changes can be merged.

amrbekhit:
I've made a new commit that modifies the tests to reflect the changes made in my original commit. Essentially, the original code used to handle QoS2 reconnects the same, regardless of whether the client is using a clean session or not, and the original tests were written based on that. My patch corrects this issue so that QoS2 reconnects are now different depending on the clean session state. I've modified the existing test and added a new one to test both of these situations.

PierreF:
I disagree with the with on loop_misc fix. The client is perfectly allowed to immediately start sending packet. From MQTT 3.1.1 spec:

> Clients are allowed to send further Control Packets immediately after sending a CONNECT Packet; Clients need not wait for a CONNACK Packet to arrive from the Server.

I'm not sure to agree on the way to fix this issue for QoS 2 full resend. I fear that in some case it could result in two delivery of the message (publish, pubrec, pubrel, server send the pubcomp but connection break before puback reach the client).
In addition to my understanding this is NOT compliant with MQTT specification which state:

> If CleanSession is set to 1, the Client and Server MUST discard any previous Session
> [...]
> The Session state in the Client consists of:
> * QoS 1 and QoS 2 messages which have been sent to the Server, but have not been completely acknowledged.

So from the standard, we must discard any QoS 1&2 message for which the publish were sent to the server in case of reconnection with clean session.
I agree that we are probably not compliant for QoS 1 packet and resend them. But going with a non-standard behavior should not cause unexpected result (QoS 2, Exactly once delivery, but could cause multiple delivery). And yes the standard means that with cleanSession=1 you may get zero delivery.

To summarize:
* the fix on loop_misc must not be needed. It's a broker bug.
* We should follow standard cleanSession=1 handling (probably with some documentation reminding that cleansession=1 means possible lost)
* If going with non-standard behavior, it should be optional, documented as such and should avoid possible multiple delivery of QoS message 2.

PR #282, opened 2018-03-22
Add bridge support

vrst37:
Mosquito broker and RSMB support a bridge connection. 

If the client establishes a bridge connection, then the broker does not send messages sent from bridge back to the bridge thus avoiding loops. It also enables proper propagation of retain flag. With this feature, now the library can be used to create bridges between two brokers. 

This implementation adds a function that can be used to enable this feature. The function must be called before the connect function (just like set user-name and password function).

The implementation adds two global constants to identify the mode being used, and if the mode is bridge mode, then when sending the connect packet, sets the MSB of protocol version bit high (indicating a bridge connection to the broker)

### DIFF
--- a/src/paho/mqtt/__init__.py
+++ b/src/paho/mqtt/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "1.4.0.dev0"
+__version__ = "1.4.1.dev0"
 
 
 class MQTTException(Exception):
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -142,6 +142,9 @@ MQTT_ERR_UNKNOWN = 13
 MQTT_ERR_ERRNO = 14
 MQTT_ERR_QUEUE_SIZE = 15
 
+MQTT_CLIENT = 0
+MQTT_BRIDGE = 1
+
 sockpair_data = b"0"
 
 class WebsocketConnectionError(ValueError):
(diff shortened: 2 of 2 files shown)


PierreF:
What is your use-case to require this feature ? I would like to be sure your use-case can't be addressed with existing feature to avoid adding unneeded feature.
Also (but I could fix it while merging) I would like to mark that this is non-standard feature and nothing is guaranteed (even Mosquitto said that loop detection is "more effective", not perfect).

vrst37:
Hi @PierreF 

I wanted to create a bridge between two MQTT brokers, filter a few messages between them and act on them before forwarding them. 

My main use case is to stop the rebroadcast of messages to the sender (i.e. if MQTT Client A sends a message XYZ, I do not want MQTT Client A (subscribed to #) to get that message. I created application side logic, but the implementation was not scalable.

I am perfectly happy marking this as a non standard feature and we should be ready to remove this feature if Mosquitto drops support for this feature.

src/paho/mqtt/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "1.4.0.dev0"
+__version__ = "1.4.1.dev0"
```
PierreF: This version should stay 1.4.0 as it's the future version number. I'll fix it after merging it.

PierreF:
It will be merged today :)
For the release, I expect to do a feature release early in June (for the annual coordinated paho release).

vrst37:
@PierreF any update on the pip release? When can we expect a release?

Issue #284, opened 2018-03-29
Paho freezes if resuming session that starts with PUBREL


rbalik:
We're working the emqtt broker which appears to resume persistent sessions at exactly the point in the QOS2 handshake where they left off. This means the broker may start the session by sending PUBREL instead of PUBLISH.
Paho doesn't reply to the PUBREL and so the session ends up getting stuck (we only have one message in flight at a time) and emqtt keeps sending the same PUBREL over and over again.

The emqtt behavior appears to be based on a very strict reading of the spec:
> If CleanSession is set to 0, the Server MUST resume communications with the Client based on state from the current Session (as identified by the Client identifier).
> The Session state in the Client consists of:
> QoS 2 messages which have been received from the Server, but have not been completely acknowledged.

mosquitto_sub handles this behavior by sending a PUBCOMP every time which seems to be a little more usable.

Here's the log output from paho:
```
INFO:root:Sending CONNECT (u1, p1, wr0, wq0, wf0, c0, k60) client_id=reubenpy2
INFO:root:Received CONNACK (1, 0)
INFO:root:Connected with result code 0
INFO:root:Session present: 1
INFO:root:Sending SUBSCRIBE (d0) [('#', 2)]
INFO:root:Received PUBREL (Mid: 32629)
INFO:root:Received SUBACK
INFO:root:subscribed with qos:(2,)
```

I put a quick patch together to change _handle_pubrel to always send a PUBCOMP back and I'll post a PR for that.

rosshettel:
We're seeing this too, any idea when the PR might get merged?

PierreF:
In you test, could you confirm that:
* you are using cleanSession=0
* you restarted the paho client and not just a reconnection (i.e. the memory state is lost)

The issue that will come from your proposed change it that some QoS 2 messages may be lost... but I agree that it may be better that no longer working at all.

rbalik:
Hi @PierreF did you have a chance to think about this?
This is causing some issues in a project I'm working on. Would be good to get that patch merged in for an update.

PierreF:
Merged. I'll update documentation to state the limitation of current implementation: since we don't have persistent store, it's possible to loss some Qos2 message on client restart.

PR #285, opened 2018-03-29
Always send back a PUBCOMP

rbalik:
See #284 

This may not be the most correct behavior but seems like the only way to stop a session from getting stuck forever in the specific case listed in the issue.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -2535,9 +2535,7 @@ class Client(object):
                         if rc != MQTT_ERR_SUCCESS:
                             return rc
 
-                    return self._send_pubcomp(mid)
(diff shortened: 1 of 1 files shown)


vrst37:
Hi, you will need to sign Eclipse contributor agreement (https://www.eclipse.org/legal/ECA.php) and have a signed-off commit (use commit -s) before this commit can be accepted.

rbalik:
Hi I signed it last night but it was after I posted the PR. Is there a way to refresh?

rbalik:
Oh I didn't do the commit -s. I'll do that soon.

rbalik:
I'm gonna do this over a better way and resubmit.

PR #286, opened 2018-04-02
Always send back a PUBCOMP so things don't get stuck

rbalik:
This is a quick patch to make sure we always respond to a PUBREL with a PUBCOMP so a session doesn't get stuck.

Signed-off-by: Reuben Balik <rsbalik@gmail.com>

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -2706,9 +2706,7 @@ class Client(object):
                         if rc != MQTT_ERR_SUCCESS:
                             return rc
 
-                    return self._send_pubcomp(mid)
-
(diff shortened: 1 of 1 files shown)


rbalik:
See #284

rbalik:
Not sure why that test is failing. All tests pass on my machine.

Also, I guess I used the wrong email address for the patch, but the signed by user is who my eclipse account is under. Anyway to fix that?

rbalik:
Fixed author. Build seems to be randomly failing.

PierreF:
I'll have a look at the issues this week-end. Thank for your patience.

Issue #288, opened 2018-04-10
TimeoutError not handled


mlhales:
https://github.com/eclipse/paho.mqtt.python/blob/e9914a759f9f5b8081d59fd65edfd18d229a399e/src/paho/mqtt/client.py#L1470

When the client is connected using [connect_async()](https://github.com/eclipse/paho.mqtt.python/blob/e9914a759f9f5b8081d59fd65edfd18d229a399e/src/paho/mqtt/client.py#L808) and then loop_start() is called which runs [loop_forever()](https://github.com/eclipse/paho.mqtt.python/blob/e9914a759f9f5b8081d59fd65edfd18d229a399e/src/paho/mqtt/client.py#L2650) in its own thread. The socket connection which happens in [reconnect()](https://github.com/eclipse/paho.mqtt.python/blob/e9914a759f9f5b8081d59fd65edfd18d229a399e/src/paho/mqtt/client.py#L895) can throw a TimeoutError which is not not caught, and cannot be caught by user code since the thread is owned by the library.

This can be fixed by catching socket.timeout in addition to socket.error, or by catching OSError.

socket.error is documented as a deprecated alias for OSError, but it does not seem to include TimeoutError which is a subclass of OSError

ncoghlan:
It looks like this has been resolved in #291, so should this issue be closed as fixed in the v1.4.0 release?

prsousa:
I'm having the same problem, but during `_packet_read`.

https://github.com/eclipse/paho.mqtt.python/blob/9782ab81fe7ee3a05e74c7f3e1d03d5611ea4be4/src/paho/mqtt/client.py#L2369-L2376
`self._sock_recv(1)` is raising a `TimeoutError` ([Errno 60] Operation timed out) and the thread created by `loop_start` crashes (without recovering).
Is this an issue of the library, or should it be handled on an upper level?

MattBrittan:
I'm going to close this because the original issue appears to have been resolved (and there is a separate issue, #694, which looks like a better place to discuss the related issue in the comment from @prsousa). This is part of a general project to clean-up issues (which should make it simpler to identify priorities going forward).

PR #289, opened 2018-04-10
Handle OSError instead of socket.error

mlhales:
Socket library docs indicated that socket.error is an alias for OSError, but TimeoutError is not caught. This means that the error handling in loop_forever does not work correctly when retry_first_connection=True

Issue is [here](https://github.com/eclipse/paho.mqtt.python/issues/288)

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -1468,7 +1468,7 @@ class Client(object):
             if self._state == mqtt_cs_connect_async:
                 try:
                     self.reconnect()
-                except (socket.error, OSError, WebsocketConnectionError):
+                except (OSError, WebsocketConnectionError):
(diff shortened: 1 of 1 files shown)


PierreF:
Looks good to me. socket.error is a subclass of OSError in Python 2.7 (tested, since I didn't find it on doc) and 3.4+ (since docs said it's the case since 3.3).
About travis-ci the error is an unstable test. I've re-run the test which now pass.

Could you however fix the IP validation check which require:
* To sign the ECA: https://www.eclipse.org/legal/ECA.php
* To commit with -s (sign-off)

Thanks.

mlhales:
I'm closing this pull request because I made of mess of my forked branch. Will re-open on a clean fork, with the proper IP validation and the change as a single commit.

PR #291, opened 2018-04-15
Catch OSError in addition to socket.error and WebsocketConnectionError

mlhales:
Socket library docs indicated that socket.error is an alias for OSError in Python 3 but TimeoutError is not caught. This means that the error handling in loop_forever does not catch TimeoutError when retry_first_connection=True leading to an unsreposive thread.

Signed-off-by: Mike Hales <mike.hales@kscape.com>

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -1468,7 +1468,7 @@ class Client(object):
             if self._state == mqtt_cs_connect_async:
                 try:
                     self.reconnect()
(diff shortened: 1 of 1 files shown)


mlhales:
[Reopened](https://github.com/eclipse/paho.mqtt.python/pull/289) with clean forked branch, ip-validation and fix that works for both python27 and python3 +

ncoghlan:
Mentioning #288 over here so GH creates a clear cross-link in both directions.

PR #294, opened 2018-04-30
Add protection around the user-defined callbacks

swails:
If a user-defined callback raises an Exception in a subscription, that
could bring down the only thread that is listening on subscriptions.

Signed-off-by: Jason Swails <jason.swails@gmail.com>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -35,7 +35,7 @@ Contents
         * `Using Callback`_
 * `Reporting bugs`_
 * `More information`_
-   
+
 
 Installation
 ------------
@@ -80,10 +80,10 @@ Getting Started
 
 Here is a very simple example that subscribes to the broker $SYS topic tree and prints out the resulting messages:
 
-::
+.. code:: python
 
     import paho.mqtt.client as mqtt
-    
+
     # The callback for when the client receives a CONNACK response from the server.
     def on_connect(client, userdata, flags, rc):
         print("Connected with result code "+str(rc))
@@ -128,7 +128,7 @@ Constructor / reinitialise
 Client()
 ''''''''
 
-::
+.. code:: python
 
     Client(client_id="", clean_session=True, userdata=None, protocol=MQTTv311, transport="tcp")
 
@@ -144,7 +144,7 @@ clean_session
     remove all information about this client when it disconnects. If ``False``,
     the client is a durable client and subscription information and queued
     messages will be retained when the client disconnects.
-    
+
     Note that a client will never discard its own outgoing messages on
     disconnect. Calling connect() or reconnect() will cause the messages to be
     resent. Use reinitialise() to reset a client to its original state.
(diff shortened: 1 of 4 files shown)


PierreF:
This PR contains lots of unrelated commit (probably due to including the merge of fixes branch into your branch).
This make the PR unreviewable, could you only submit commit related to this change ?

While I think this change it a good move, this means it add a way to loss a message:
* previously an error on on_message delivery broke the consumer but didn't acknowledged the message
* with such protection, an error on on_message delivery will still acknowledged the message even if the error probably means the message was not fully processed.

I think this behavior may worse a note on documentation.

swails:
This PR contains *only* the commit related to the topic and the merge of the fixes branch -- nothing else.  Which means it's a few try-except's, and everything else is from fixes, with manual conflict resolution.

Certainly it's harder to review than a small change, but "unreviewable" is a bit of a stretch.  But I needed to do this -- I needed to actually test my changes in the context of (what I interpreted to be) the current state of the *actual* code that would be released.  That you think these changes are unreviewable to me is just a validation of #295.

It took me 30-60 minutes to make a 5 minute change.  If you want to look at just the proposed changes for review, look at [only changeset I created for this feature](https://github.com/eclipse/paho.mqtt.python/pull/294/commits/a311fdf0f7d00c220a5af6b97646e2b62a13310e).  I'm not reversing the fixes merge I did to make this PR easier to review.

swails:
Feel free to reopen if you actually want this merged, but contributing to this project is increasingly becoming an exercise in frustration in which I'm no longer willing to participate.

PierreF:
I didn't wanted to be rude in my previous comment. What I wanted to means is that PR should be kept minimal (why the merge or fixes was needed ?) to ease review and later merge (I had too many issue with PR including unrelated change which cause harder merge later).
Yes, some those issue may be fixed by easier branching model but I don't have a solution yet for this point.

jamesmyatt:
Maybe someone can merge fixes in to develop first.

swails:
I wasn't trying to imply you were being rude, but I do want to convey my frustration with contributing here that is a direct result of the branching model (not at all the conduct of you or anyone else).

As to "why the merge of fixes was needed", I answered that in my post:

> But I needed to do this -- I needed to actually test my changes in the context of (what I interpreted to be) the current state of the actual code that would be released.

I was trying to debug an issue in one of my programs that subscribed and unsubscribed repeatedly by many threads.  If Thread-1 goes down due to an exception in my subscription callback (the one responsible for effectively calling every subscription callback), it crashes everything, because subscription callbacks stop getting called and my program was simply waiting on callbacks that were never going to happen.  I was also leaking subscriptions, and some of the changes in fixes looked applicable to what I was seeing (particularly with the mids).

So I needed what was in develop, fixes, and the changes I was proposing.  I was frustrated at having to do a manual conflict resolution on fixes and develop, but I had to do it as part of my investigation.  However, while I've done this for PRs here in the past, I'm unwilling to unravel my merge work to pull out the try-except blocks I added this time (since they're all right next to a `with self._in_callback` -> `with self._in_callback_mutex` change requiring manual conflict resolution).  If that means nobody is willing to review and merge this PR, I hold no grudge about that.  That's why I closed this PR.  You're welcome to reopen and review.  I'll address review comments, but other than that I'm not pursuing this any further.

PierreF:
Thanks for your detailed feedback on what and why you had to do extra work for testing this PR.

I reopen to remember to take the commit and merge it.

PierreF:
Cherry-picked the commit into develop branch.

simonvanderveldt:
@PierreF Any chance this could be reverted? Handling exceptions should be up to the user + catching `Exception` is generally not a thing one should be doing, definitely not in a library

I've been scratching my head why I'm not seeing any exceptions, then looked at paho code and found this PR. Not really the principe of least surprise at work here :)

PR #299, opened 2018-05-11
Allow insecure TLS option with publish.single and publish.multiple

LyleScott:
This is more of a starting point, but this is what I've been using in some projects of mine for various reasons. I've found the publish.single/publish.multiple helpers very convenient in strict use cases but the functions don't seem to support "insecure" TLS.  I've tried to keep the same terminology as `client.Client` in my solution.

1) Is there interest in carrying this forward?
2) I don't see any tests I can currently hook in to. Did I overlook one?

### DIFF
--- a/src/paho/mqtt/publish.py
+++ b/src/paho/mqtt/publish.py
@@ -109,7 +109,7 @@ def multiple(msgs, hostname="localhost", port=1883, client_id="", keepalive=60,
     tls : a dict containing TLS configuration parameters for the client:
           dict = {'ca_certs':"<ca_certs>", 'certfile':"<certfile>",
           'keyfile':"<keyfile>", 'tls_version':"<tls_version>",
-          'ciphers':"<ciphers">}
+          'ciphers':"<ciphers">, 'insecure':"<bool>"}
           ca_certs is required, all other parameters are optional and will
           default to None if not provided, which results in the client using
           the default behaviour - see the paho.mqtt.client documentation.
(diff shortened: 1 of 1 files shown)


PierreF:
Thanks for your patience.

1. Yes, this is a useful addition that was missing
2. SSL connection test are done under the "test" folder (and not under "tests" folder. All tests should be moved to "tests" folder but not yet done).

Before I could merge this PR, could you sign the Eclipse Contributor Agreement as described in https://github.com/eclipse/paho.mqtt.python/blob/master/CONTRIBUTING.md ? In short it means:
* Go to https://accounts.eclipse.org/user/eca to sign the ECA (Eclipse Contributor Agreement)
* Add "Signed-Off" to your commits

LyleScott:
I will do my best to follow up with a test.

src/paho/mqtt/publish.py
```diff
@@ -144,7 +144,12 @@ def multiple(msgs, hostname="localhost", port=1883, client_id="", keepalive=60,
 
     if tls is not None:
         if isinstance(tls, dict):
+            insecure = tls.pop('insecure', None)
             client.tls_set(**tls)
+            if insecure is True:
```
kryskool: remove `is True`because insecure is a boolean

PierreF:
Cherry-picked the commit and incorporated suggestion by @kryskool. Thanks both of you for your contribution.

PR #310, opened 2018-07-04
Republishing qos2 messages and removing incomplete received messages(_in_messages) when clean_session=true

tarunw07:
When clean_session=True, Paho client still sends the PUBREC, PUBLREL, etc packages and I think that should not happen. I have fixed it by completely removing _in_messages and changing the status of _out_messages (qos2) messages to mqtt_ms_publish.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -2172,7 +2172,13 @@ class Client(object):
             self._pack_str16(packet, t)
             packet.append(q)
 
-        self._easy_log(MQTT_LOG_DEBUG, "Sending SUBSCRIBE (d%d) %s", dup, topics)
+        self._easy_log(
+            MQTT_LOG_DEBUG,
+            "Sending SUBSCRIBE (d%d, m%d) %s",
+            dup,
+            local_mid,
+            topics,
+        )
         return (self._packet_queue(command, packet, local_mid, 1), local_mid)
 
     def _send_unsubscribe(self, dup, topics):
@@ -2190,7 +2196,13 @@ class Client(object):
             self._pack_str16(packet, t)
 
         # topics_repr = ", ".join("'"+topic.decode('utf8')+"'" for topic in topics)
-        self._easy_log(MQTT_LOG_DEBUG, "Sending UNSUBSCRIBE (d%d) %s", dup, topics)
+        self._easy_log(
+            MQTT_LOG_DEBUG,
+            "Sending UNSUBSCRIBE (d%d, m%d) %s",
+            dup,
+            local_mid,
+            topics,
+        )
         return (self._packet_queue(command, packet, local_mid, 1), local_mid)
 
     def _message_retry_check_actual(self, messages, mutex):
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -555,6 +555,7 @@ class Client(object):
         self._out_message_mutex = threading.RLock()
         self._in_message_mutex = threading.Lock()
         self._reconnect_delay_mutex = threading.Lock()
+        self._mid_generate_mutex = threading.Lock()
         self._thread = None
         self._thread_terminate = False
         self._ssl = False
@@ -1966,10 +1967,11 @@ class Client(object):
                             self.on_disconnect(self, self._userdata, rc)
 
     def _mid_generate(self):
-        self._last_mid += 1
-        if self._last_mid == 65536:
-            self._last_mid = 1
-        return self._last_mid
+        with self._mid_generate_mutex:
+            self._last_mid += 1
+            if self._last_mid == 65536:
+                self._last_mid = 1
+            return self._last_mid
 
     @staticmethod
     def _topic_wildcard_len_check(topic):
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -2249,18 +2249,26 @@ class Client(object):
                         m.state = mqtt_ms_publish
                     elif m.qos == 2:
                         # self._inflight_messages = self._inflight_messages + 1
-                        if m.state == mqtt_ms_wait_for_pubcomp:
-                            m.state = mqtt_ms_resend_pubrel
-                            m.dup = True
-                        else:
-                            if m.state == mqtt_ms_wait_for_pubrec:
+                        if self._clean_session is True:
+                            if m.state != mqtt_ms_publish:
                                 m.dup = True
                             m.state = mqtt_ms_publish
+                        else:    
+                            if m.state == mqtt_ms_wait_for_pubcomp:
+                                m.state = mqtt_ms_resend_pubrel
+                                m.dup = True
+                            else:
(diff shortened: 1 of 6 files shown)


PierreF:
This PR is similar to #281.
It also have the possible issue with multiple delivery of QoS 2 message (publish, pubrec, pubrel, server send the pubcomp but connection break before puback reach the client; client result as publish).
Once more this is not compliant by my understanding of standard:

> If CleanSession is set to 1, the Client and Server MUST discard any previous Session
> [...]
>    The Session state in the Client consists of:
>        QoS 1 and QoS 2 messages which have been sent to the Server, but have not been completely acknowledged.

Once publish packet were sent, in case of re-connection with clean_session=1 we need to discard the message completely. As said in #281 this may worse a note in the README.

src/paho/mqtt/client.py
```diff
@@ -2249,18 +2249,26 @@ def _messages_reconnect_reset_out(self):
                         m.state = mqtt_ms_publish
                     elif m.qos == 2:
                         # self._inflight_messages = self._inflight_messages + 1
-                        if m.state == mqtt_ms_wait_for_pubcomp:
-                            m.state = mqtt_ms_resend_pubrel
-                            m.dup = True
-                        else:
-                            if m.state == mqtt_ms_wait_for_pubrec:
+                        if self._clean_session is True:
```
PierreF: Drop the "is True"
```
if self.self_session:
   [...]
```

src/paho/mqtt/client.py
```diff
@@ -2249,18 +2249,26 @@ def _messages_reconnect_reset_out(self):
...
+                            if m.state != mqtt_ms_publish:
                                 m.dup = True
```
PierreF: As said in PR global comment, with clean_session=True we should discard any message for which the publish packet was send.

tarunw07: If I drop the "is True" some test cases(03-publish-c2b-qos1-disconnect and maybe more) fail. Why are you passing clean_session=-1 in test cases?
````
run = -1`
mqttc = mqtt.Client("publish-qos1-test", run)
````

Should I update the _messages_reconnect_reset_out to clear the _out_messages when clean_session=True and what should I do about testcases?

Comment (PierreF) [reply src/paho/mqtt/client.py]: It's expected that this test fail :)
I don't really know why there is this "run = -1" at all... in all test I looked that I looked for it's a constant used in a "while run == -1" (which is thus equivalent to while True).
In this test is used to set clean_session=-1 (equivalent to clean_session=True)... which make the test wrong since this test expect a session to persist a disconnection or timeout.

Comment (tarunw07) [reply src/paho/mqtt/client.py]: Ok cool! I see you have updated the test cases. I have modified the pull request to completely remove _out_messages messages when clean_session=True.

File: src/paho/mqtt/client.py [hunk 3]
```diff
@@ -2237,6 +2237,11 @@ def _message_retry_check(self):
     def _messages_reconnect_reset_out(self):
         with self._out_message_mutex:
             self._inflight_messages = 0
+            
+            if self._clean_session:
+                self._out_messages = []
```
PierreF: I don't think the standard impose to discard *all* messages. We must only discard message for which the publish packet was sent.
_out_messages contains all messages that will be send or are currently being send (for QoS > 0).
Excepted if i'm missing a part of the standard, I think we should keep not yet send messages. Without this, code like this one will start to fail:
```
client = mqtt.Client()
msg = client.publish('topic', qos=1)  # call publish *before* connection
client.connect("localhost", 1883, 60)
client.loop_start()
msg.wait_for_publish()  # this will hang forever if we drop all _out_messages
```

tarunw07: Okay....That will work as well!!
Is this the correct way to implement this approach? Is it possible that message state(m.state) is mqtt_ms_queued(because we have to keep those messages as well)? I can't think of any case where message state is mqtt_ms_queued.

     def _messages_reconnect_reset_out(self):
        with self._out_message_mutex:
            self._inflight_messages = 0
            
            if self._clean_session:
                for m in self._out_messages:
                    m.timestamp = 0
                    if m.state != mqtt_ms_publish:
                        self._out_messages.pop(self._out_messages.index(m))
                return
    
            for m in self._out_messages:
                m.timestamp = 0
                if self._max_inflight_messages == 0 or self._inflight_messages < self._max_inflight_messages:
                    if m.qos == 0:
                        m.state = mqtt_ms_publish
                    elif m.qos == 1:
                        # self._inflight_messages = self._inflight_messages + 1
                        if m.state == mqtt_ms_wait_for_puback:
                            m.dup = True
                        m.state = mqtt_ms_publish
                    elif m.qos == 2:
                        # self._inflight_messages = self._inflight_messages + 1                           
                        if m.state == mqtt_ms_wait_for_pubcomp:
                            m.state = mqtt_ms_resend_pubrel
                            m.dup = True
                        else:
                            if m.state == mqtt_ms_wait_for_pubrec:
                                m.dup = True
                            m.state = mqtt_ms_publish
                else:
                    m.state = mqtt_ms_queued

PierreF: Yes I think this is a good solution. Would should also keep mqtt_ms_queued state.
mqtt_ms_publish happen when trying to send Qos > 0 and connection is not established
mqtt_ms_queued happen when we reach max_inflight_message.

In both case since publish packet were not sent, we can keep the message for the next session.

PierreF:
Thanks, I think that this make us compliant with MQTT.
But I forget that in our documentation say:
> [about] clean_session
> [...]
> Note that a client will never discard its own outgoing messages on disconnect. Calling connect() or reconnect() will cause the messages to be resent. Use reinitialise() to reset a client to its original state.

Therefor discarding message like we does will be a breaking change (and a lost of message when documentation said we kept them). So I start to be mitigated about how to handle it.
I can think of the following solution:
* Keep only your first commit, we are not fully compliant (especially it's possible to get duplicated QoS 2) but we don't hang with a compliant broker and keep the old behavior
* Make such breaking change (should it be with a release 2.0 ?). But there is few issue, esp. the publish helper which are using clean_session=True but we expect them to successfully sent the message.
* Have a "strict_mode" boolean to select which of the two above mode should be used. Default to strict_mode=False to keep previous behavior. It may allow soother migration to 2nd solution.

I'm not using client with clean_session=True, so I don't really known what user expect in such case.

tarunw07:
I think we should go with my first commit and mention in the docs that it's possible to get duplicated QoS 2 message if there is disconnection when clean_session is True. And to avoid this, a user can either set clean_session=False or use reinitialise() method which will discard all the messages yet to be published.

The problem with the current solution is that it is not deterministic(because some messages are discarded) and that can cause some trouble I guess.

The best option would've been to remove all the messages but that will cause wait_for_publish() to hang forever and that's why we can't go with this option.

What do you think?

PierreF:
I think this is the best compromise for now. I will take your first commit and add a note on possible duplicated QoS 2.

Issue #20, opened 2016-02-04
Python client timeout if behind a HTTP proxy


jpwsutton:
migrated from Bugzilla [#457256](https://bugs.eclipse.org/bugs/show_bug.cgi?id=457256)
status CLOSED severity *normal* in component *MQTT-Python* for *1.1*
Reported in version *v0.5* on platform *PC*
Assigned to: Roger Light

On 2015-01-12 10:32:39 -0500, DD G wrote:

> Unable to use the Python client on a machine that sits behind a corporate HTTP proxy.  There doesn't seem to be a way to configure the proxy details.
> 
> I was able to make mosquitto_sub command line tool work using proxychains as a workaround.  However a similar approach failed with the Python script.
> 
> -----------------------------------------------------------
> #!/usr/bin/env python 
> 
> import paho.mqtt.client as mqtt
> import time
> 
> client = mqtt.Client()
> client.connect("test.mosquitto.org", 1883, 60)
> temp = 1
> 
> client.loop_start()
> while True:
>         temp = temp + 1
>         client.publish("house/temp/xxxx123", temp, retain=True)
>         time.sleep(10)
> 
> client.loop_forever()
> 
> ----------------------------------------------------------------
> 
> when invoked without proxychains:
> 
> >python test.py 
> 
> python broker-sys.py 
> Traceback (most recent call last):
>   File "tests.py", line 31, in <module>
>     client.connect("test.mosquitto.org", 1883, 60)
>   File "/usr/local/lib/python2.7/dist-packages/paho/mqtt/client.py", line 600, in connect
>     return self.reconnect()
>   File "/usr/local/lib/python2.7/dist-packages/paho/mqtt/client.py", line 722, in reconnect
>     self._sock = socket.create_connection((self._host, self._port), source_address=(self._bind_address, 0))
>   File "/usr/lib/python2.7/socket.py", line 571, in create_connection
>     raise err
> socket.error: [Errno 101] Network is unreachable
> 
> with proxychains (proxy IP changed to 99.99.99.99):
> 
> >proxychains python test.py 
> ProxyChains-3.1 (http://proxychains.sf.net)
> |S-chain|-<>-99.99.99.99:8080-<><>-127.0.0.1:54202-<--denied
> Traceback (most recent call last):
>   File "test.py", line 26, in <module>
>     client = mqtt.Client()
>   File "/usr/local/lib/python2.7/dist-packages/paho/mqtt/client.py", line 410, in __init__
[...]

PR #315, opened 2018-07-25
Added proxy support

gtmtg:
Hey! I made a couple changes that I needed to use Paho for my project and that I think will be helpful to other developers as well.

1) Added support for establishing MQTT connections behind HTTP and SOCKS proxies through the [PySocks](https://github.com/Anorov/PySocks) library. This has been requested in Paho by a number of people; see #20, [here](https://stackoverflow.com/questions/27834468/how-to-add-proxy-settings-in-paho-mqtt), [here](https://stackoverflow.com/questions/43569068/python-paho-mqtt-with-socks-proxy), [here](https://www.eclipse.org/forums/index.php/t/720281/), and [here (for Java)](https://stackoverflow.com/questions/30934260/connecting-to-external-broker-through-restricted-network-using-paho-java-client) for a few requests for this feature. I needed this to use Paho on my company's servers as well.

2) Cleaned up error handling in a couple places. This is very minor, but instead of catching errors with a plain `except:`, as the original code did, I'm now catching `except Exception:`, which doesn't hide system errors (like running out of memory) under the surface. See [here](https://stackoverflow.com/questions/18982610/difference-between-except-and-except-exception-as-e-in-python) for more details.

Thanks so much for an awesome library!


src/paho/mqtt/client.py
```diff
@@ -2671,6 +2694,89 @@ def _reconnect_wait(self):
...
+                "proxy_port": socks_default[2],
+                "proxy_rdns": socks_default[3],
+                "proxy_username": socks_default[4],
+                "proxy_password": socks_default[5]
+            })
+
+        # Finally, look at environment variables ending in _proxy as long as the
+        # host we're trying to connect to isn't listed under the no_proxy
+        # environment variable (matches built-in module urllib's behavior)
+        if not urllib_dot_request.proxy_bypass(self._host):
```
PierreF: I can't find documentation on proxy_bypass. I'm unsure how safe it's to rely this undocumented method.
I'm especially thinking about:
* Since it's not documented, I don't known from which version of Python is appear
* Since it's not documented, why it can't go away in next Python version ?

Maybe at least adding an hasattr() would help (on the other hand it means that if the function disappear this feature will be silently ignored)

src/paho/mqtt/client.py
```diff
@@ -2671,6 +2694,89 @@ def _reconnect_wait(self):
...
+                return bool(p.get("proxy_type"))
+            elif isinstance(p, (list, tuple)):
+                return len(p) == 6 and bool(p[0])
+            else:
+                return False
+
+        proxies = []  # proxy servers to use in order of preference
+
+        if proxy_is_valid(self._proxy):
+            proxies.append(self._proxy)
```
PierreF: If the user *explicitly* give a proxy parameter to the connect() method, shouldn't we only try this proxy ?

In general shouldn't we only try the first proxy ? We could also wonder why using environment variable if user explicitly called socks.set_default_proxy().

src/paho/mqtt/client.py
```diff
@@ -2671,6 +2694,89 @@ def _reconnect_wait(self):
...
+                "proxy_username": socks_default[4],
+                "proxy_password": socks_default[5]
+            })
+
+        # Finally, look at environment variables ending in _proxy as long as the
+        # host we're trying to connect to isn't listed under the no_proxy
+        # environment variable (matches built-in module urllib's behavior)
+        if not urllib_dot_request.proxy_bypass(self._host):
+            env_proxies = urllib_dot_request.getproxies()
+            for proxy in env_proxies.values():
```
PierreF: If I understand well, this will catch an "http_proxy" or "https_proxy" to proxy MQTT protocol. Shouldn't we use an "mqtt_proxy" variable ?

The other question I have is how to disable proxying. Let's imaging my program doing both http request and use paho-mqtt. How to tell it to use the proxy for http but not mqtt ?

src/paho/mqtt/client.py
gtmtg: Good point. How about:

```python
if not (hasattr(urllib_dot_request, "proxy_bypass") and urllib_dot_request.proxy_bypass(self._host)):
```

I think this makes sense as a fallback if the function disappears—since the function tells us which connections we *shouldn't* proxy, if the function doesn't exist, we proxy connections using the supplied connections by default.

src/paho/mqtt/client.py
gtmtg: Great catches for both of these. I will change the code to use an `mqtt_proxy` variable only. For the second question, what if we add a `no_proxy` flag to `connect()` and related methods that defaults to True but, if set to False, bypasses all proxies for MQTT connections?

gtmtg:
Also, question: where is the best place to document usage of the `mqtt_proxy` environment variable?

gtmtg:
@PierreF any update/feedback here?

src/paho/mqtt/client.py
```diff
@@ -2671,6 +2694,83 @@ def _reconnect_wait(self):
             time.sleep(min(remaining, 1))
             remaining = target_time - time_func()
 
+    def _get_proxy(self):
+        def proxy_is_valid(p):
+            if isinstance(p, dict):
+                return bool(p.get("proxy_type"))
+            elif isinstance(p, (list, tuple)):
```
jamesmyatt: Why is list/tuple allowed?

src/paho/mqtt/client.py
```diff
@@ -993,10 +1019,7 @@ def loop(self, timeout=1.0, max_packets=1):
             # Can occur if we just reconnected but rlist/wlist contain a -1 for
             # some reason.
             return MQTT_ERR_CONN_LOST
-        except KeyboardInterrupt:
-            # Allow ^C to interrupt
-            raise
-        except:
+        except Exception:  # Still allows ^C (KeyboardInterrupt) to interrupt
```
jamesmyatt: I would clarify this as "Note that KeyboardInterrupt, etc. are not derived from Exception`

jamesmyatt:
Superficial comments from an interested bystander:

* I think that if you're going to implement this, then it should use the same pattern as the SSL configuration. i.e. rather than passing the parameters through the `connect*` methods, use a separate `proxy_set` method, etc.

* I would try to make PySocks an optional dependency. So far, it looks like there's been an objective to keep install dependencies to the standard library, and very few users will need to use this proxy support. Look at the `extras_require` setting in setuptools.

PierreF:
Thanks @jamesmyatt for your review and good comments.
Speaking of PySocks dependency, adding a new dependencies will require a IP review for which I need to find what the process is.

src/paho/mqtt/client.py
gtmtg: PySocks' `socks.get_default_proxy()` method returns the configured proxy as a tuple. Also, I might not be understanding properly, but how would `**` unpack a list/tuple?

gtmtg:
@jamesmyatt Thanks for the review — really appreciate it :) Made the changes! How does the latest version look?

@PierreF Awesome — let me know if I need to do anything to help with the IP review. Does everything else look good?

gtmtg:
@PierreF any updates here?

harshadk-mcafee:
Hello, 
We need to add proxy support in Paho-MQTT python for our project. Are there any update on this?

gtmtg:
@harshadk-mcafee Feel free to use my fork, I'm no longer using it but I think it should work fine!

@PierreF any updates on getting this merged upstream?

dlxneamtu:
Hi,

I'd like to use this, but can you please give me an example of how exactly I should call proxy_set() function in order to use an HTTP proxy with hostname=20.20.20.20:80, no username and passoword required? 

Thanks

src/paho/mqtt/client.py
jamesmyatt: I think I meant `*`.

src/paho/mqtt/client.py
```diff
@@ -23,10 +23,24 @@
...
+    pass
+
+socks = None
+try:
+    import socks
+
+    # Python 3
+    from urllib import request as urllib_dot_request
+    from urllib import parse as urllib_dot_parse
+except ImportError:
```
jamesmyatt: This ImportError catches 2 different kinds: one is for socks and the other is for urllib. I'd make them 2 separate try statements. I think it's OK to do the urllib imports even if socks is None. Alternatively, put the second try in the else clause of the first.

src/paho/mqtt/client.py
```diff
@@ -2671,6 +2705,84 @@ def _reconnect_wait(self):
...
+                        "proxy_addr": parts.hostname,
+                        "proxy_port": parts.port
+                    }
+                    return proxy
+
+        # Finally, check if the user has monkeypatched the PySocks library with
+        # a default proxy
+        socks_default = socks.get_default_proxy()
+        if self._proxy_is_valid(socks_default):
+            proxy = {
```
jamesmyatt: Better using `zip`

PierreF:
Sorry for the delay on this PR. I've done the IP process needed for the new dependencies which resulted in "Works with" dependencies.
So everything is good on my side, could you just fix the merge conflict and remark made by @jamesmyatt  (thanks for the review) and I will merge it.

gtmtg:
@PierreF Awesome, should be good now on the code side! I accidentally committed with my personal email (mihir.garimella@gmail.com) instead of my work email (mtg@fb.com), but I just signed the ECA with my personal email as well. It's complaining because one of the commits authored by mihir.garimella@gmail.com had a footer note that it was signed by mtg@fb.com (I amended because I thought that would fix the error), but now git is getting really messed up when I try to go back and rebase to fix the issue because I merged in upstream commits. Anything you recommend?

PierreF:
Given that you signed the ECA with both emails, I think it good for eclipse.
If it's good for you to commit with your personal email, I can force the merge like this. If not and since there is a merge (so you can't rebase), you need to reset your branch to the state before the merge to amend the commit, then re-do the merge.

youyuge34:
@gtmtg 
```
import socks
.........
client.proxy_set({'proxy_tpe':socks.HTTP,'proxy_addr':'219.219.113.223:81'})
TypeError: proxy_set() takes 1 positional argument but 2 were given
```
Why it is wrong?

gtmtg:
@youyuge34 Two things:

1) I think you have a typo; change `proxy_tpe` to `proxy_type`
2) They're passed as kwargs in Python (that's what the `**` means), so no need to pass as a dictionary.

Try this:

`client.proxy_set(proxy_type=socks.HTTP, proxy_addr='219.219.113.223:81')

gtmtg:
@PierreF personal email is fine, you can go ahead with the merge! Thanks for all the work on this (and thanks @jamesmyatt for the comments!)

antechrestos:
Great work. When will it be released?

thibleroy:
> 
> 
> @youyuge34 Two things:
> 
>     1. I think you have a typo; change `proxy_tpe` to `proxy_type`
> 
>     2. They're passed as kwargs in Python (that's what the `**` means), so no need to pass as a dictionary.
> 
> 
> Try this:
> 
> `client.proxy_set(proxy_type=socks.HTTP, proxy_addr='219.219.113.223:81')`

It was a year ago but I have to give my answer... I had to give the proxy port using the *proxy_port* parameter to get this morking

chrismaes87:
for future visitors to this page:

`client.proxy_set(proxy_type=socks.HTTP, proxy_addr='1.2.3.4', proxy_port=4231)`

PR #317, opened 2018-07-27
Travis CI: Add Python 3.7 and flake8 tests

cclauss:
* Add Python 3.7 to Travis CI and tox in alignment with travis-ci/travis-ci#9069
* Add [flake8](http://flake8.pycqa.org) tests to look for Python syntax errors and undefined names
    * Add __import os__ in aws_iot.py
* Add the Python 3.7 trove classifier to setup.py

__E901,E999,F821,F822,F823__ are the "showstopper" flake8 issues that can halt the runtime with a SyntaxError, NameError, etc.  Most other flake8 issues are merely "style violations" -- useful for readability but they do not effect runtime safety.
* F821: undefined name `name`
* F822: undefined name `name` in `__all__`
* F823: local variable `name` referenced before assignment
* E901: SyntaxError or IndentationError
* E999: SyntaxError -- failed to compile a file into an Abstract Syntax Tree


PierreF:
Thanks for your contribution.
Can can ignore the "connection reset error"... it's a false positive that need a fix, but current it mostly only happen on travis (at least I can hardly reproduce it on my laptop :().

I'm trying to cherry-pick you commit on develop branch, on which I have:
* no issue with pylama (unsure why... It Python 3.7 from Ubuntu bionic, which seems to be a beta release... maybe that why).
* but flake8 fail on Python 2.7 since it try to validate the syntax of https://github.com/eclipse/paho.mqtt.python/blob/develop/examples/loop_asyncio.py.

Do you have any suggestion for the second point ? I think of excluding "example" directory but if you have a better solution that would allow linting of other example that would be great.

CAM-Gerlach:
Its almost 2020 now, and Python 3.8 has been released, with Python 3.7 being the default for Debian/Raspbian 10 (Buster), Fedora, Arch, and other mainstream distros. Is there any progress on declaring proper Python 3.7 compatibility? I'd like to use MQTT with a remote lightning sensor network I'm in the final stages of developing, but given everything else (clients and server) is running Python 3.7 and the various packages I've built for this all written and tested on the same, I would strongly hesitate to use a library that is not tested and officially compatible with a modern version of Python not only for the obvious reasons but due to the resulting concerns about the project's long-term health and maintenance given the same. Could someone please clarify the situation with this PR and Python 3.7 support? Thanks.

cclauss:
Closed in favor of #437

PR #374, opened 2019-04-10
Leave write_loop() call to socket_register_write callback if present.

### LINKED ISSUE #361: on_publish() called before publish() completes
While trying to write an asyncio-friendly wrapper, I have this code:
```python
async def publish(self, *args, **kwargs):
    future = self.loop.create_future()
    result, mid = super().publish(*args, **kwargs)
    if result != mqtt.MQTT_ERR_SUCCESS:
        raise MQTTError(result)
    self.__in_flight[mid] = future
    return await future

def on_publish(self, client, userdata, mid):
    future = self.__in_flight.pop(mid)
    future.set_result(None)
```
The problem is that for `qos=0` messages, `on_publish` is being called somewhere from within `publish`, before it has returned the message id to us, so the above fails.

Is it possible to tell based on the `publish` result whether `publish` has completed all required actions succesfully (even when `qos>0`) or there is still work to be done and we have to wait for `on_publish`?

And how would one tell that a publish failed? - but feel free to redirect me to the protocol spec on this one if I'm missing something.

jeeger:
Instead of calling write_loop() directly when using an external event
loop, check whether the socket_register_write callback is set, and use
it if it is available.

This fixes #361, since on_publish() is now called from the
socket_register_write_callback that runs in a different
thread/coroutine instead of by the publish() function itself.

Note that this has only been tested with the asyncio external event loop, not with other external event loops.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -2574,7 +2574,9 @@ class Client(object):
             if err.errno != EAGAIN:
                 raise
 
-        if self._thread is None:
+        # If we have an external event loop registered, use that instead
+        # of calling lling loop_write() directly.
+        if self._thread is None and self.socket_register_write is None:
             if self._in_callback_mutex.acquire(False):
                 self._in_callback_mutex.release()
                 return self.loop_write()


ralight:
Thank you, this makes a lot of sense. Sorry for the absolute age to look at it.

ralight:
I merged to master by mistake, I've removed that and merged to 1.6.x instead but it doesn't show here.

PR #389, opened 2019-05-27
adding bind_port functionality for client

naveenrobo:
Adding bind_port functionality to client so that it will be possible to make the client open a connection on given port or any when nothing is specified. It will be useful when there is firewall restriction where incomings are blocked. Using static port for client will allow user to open firewall for that particular port.


naveenrobo:
Let me know what needs to be done for the legal agreements.

naveenrobo:
will be making another commit with signed off message.

Issue #402, opened 2019-08-05
proxy should be configurable from single/multiple operations


antechrestos:
With the great work of #315 , it would be great to have the possibility to specify a proxy while calling `single`or `multiple` operation

PR #403, opened 2019-08-05
Add proxy support for 'single' and 'multiple' operations

antechrestos:
This change attempts to solve issue  #402


antechrestos:
I don't know why tests are failing due to a missing `six` import

antechrestos:
After some investigation, it seems that the pipeline is broken for python 3.5 and 3.6 as six is not installed anymore by tox for these python versions...

PierreF:
The travis error is like to be the same that the one i've fixed in commit c3c8598 (cf https://github.com/eclipse/paho.mqtt.python/pull/334#issuecomment-512695890).
Tox is ok on local, i'm merging and travis will re-check on master with commit c3c8598 included.

Thank for your contribution.

antechrestos:
Thank you.

PR #407, opened 2019-08-27
Try reconnect to socket on Broken Pipe error

ninjaniko:
Signed-off-by: Nikolai Remfeldt

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -630,6 +630,12 @@ class Client(object):
                 raise WouldBlockError()
             if err.errno == EAGAIN:
                 raise WouldBlockError()
+            if err.errno == errno.EPIPE:
+                try:
+                    self.reconnect()
+                    return self._sock.recv(bufsize)
+                except:
+                    raise
             raise
 
     def _sock_send(self, buf):
@@ -644,6 +650,12 @@ class Client(object):
             if err.errno == EAGAIN:
                 self._call_socket_register_write()
                 raise WouldBlockError()
+            if err.errno == errno.EPIPE:
+                try:
+                    self.reconnect()
+                    return self._sock.send(buf)
+                except:
+                    raise
             raise
 
     def _sock_close(self):


ninjaniko:
I'm not a member of the Paho project.
Looks like build failure is equal to that of previous builds.

PierreF:
If we reconnect and retry sending bytes as you do in the PR, it may result in invalid MQTT data.
For example, while receiving large MQTT message, multiple _sock_recv will be called (because TCP socket recv() will return with data already arrived). If the broken pipe happen in the middle of that MQTT message and we reconnect + resume the _sock_recv, the called of _sock_recv will assume to get the next part of the message but will instead get the beginning of a message.

Anyway, why do you need another re-connection mechanism ? On any socket error, the reconnection should be already handled (e.g. by loop/loop_forever/loop_start).

ninjaniko:
Hi PierreF.
I notice it's handled by loop forever but if I e.g. set keepalive timer to 5 minutes and do not issue a "write to topic" command for at least every 5 minutes, then next time I try to do it I get a broken pipe error. Inserting the reconnect command solves my problems without any issues so far.

PierreF:
Are you using loop_forever/loop_start ? If you are using loop() are you calling loop() at least every 5 minutes ? (in fact, if using loop(), you should keep calling loop() as soon as possible).

loop() (which is called by loop_forever/loop_start) will perform background tasks like resent message, send large message that need multiple TCP send(), receive message, and handle keepalive. So you should not just call loop() when sending message.

ralight:
I've done some testing and as far as I can tell this only comes about if the client doesn't communicate with the broker in the keepalive interval. Dealing with keepalive is handled automatically by calling `loop_forever`, `loop_start`, or by calling `loop` or `loop_misc` at regular intervals. I don't believe this PR does anything beyond what is already available so I'm going to close it. If you disagree, please feel free to reopen it and give some example code where the existing code fails so we can better see what is happening.

Thanks for taking the time to look it.

PR #411, opened 2019-09-26
add method to check current connection status

naveenrobo:
Add method(isConnected) to check the current status of client connection with broker. Returns True/False.

Signed-off-by: Naveen Sakthivel <naveenrobotics@live.in>


naveenrobo:
I have the Signed-off-by footer. But i donno why the validation fails. Can someone assist me with this.

PR #414, opened 2019-09-26
Add method(isConnected) to check the current status of client

naveenrobo:
Add method(isConnected) to check the current status of client connection with broker. Returns True/False.

Signed-off-by: Naveen Sakthivel <naveenrobotics@live.in>

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -1345,6 +1345,14 @@ class Client(object):
         """
         self._client_mode = MQTT_BRIDGE
 
+    def isConnected(self):
+        """Returns the current status of the connection
+
+        True if connection exists
+        False if connection is closed
+        """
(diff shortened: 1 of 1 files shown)


src/paho/mqtt/client.py
```diff
@@ -1345,6 +1345,14 @@ def enable_bridge_mode(self):
         """
         self._client_mode = MQTT_BRIDGE
 
+    def isConnected(self):
```
PierreF: Could you just rename this method "is_connected" ? Python style (PEP-8) recommends using snake case for method name.

naveenrobo: I have done the changes as requested. Will follow Python styling going forward. Thanks.

tenzlobs21:
Noice

PR #421, opened 2019-09-30
add timeout for connection to the broker

ruoshui1314:


ruoshui1314:
Someone would like to tell me where is wrong.I create account and add sign off when push commit.
This pull request is important for us.So I want to pull it to the master branch.
Thank you very much.

PR #441, opened 2019-11-05
fixed the rc parameter MQTT_ERR_CONN_LOST to be passed instead of 1 t…

nawab36085:
…o user callback function on_disonncect

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -1635,7 +1635,7 @@ class Client(object):
             if self._state == mqtt_cs_disconnecting:
                 rc = MQTT_ERR_SUCCESS
             else:
-                rc = 1
+                rc = MQTT_ERR_CONN_LOST
 
             with self._callback_mutex:
                 if self.on_disconnect:


denravonska:
Is there a reason why this is still open, or why there has been no reaction to it for 4 months?

karlp:
The documentation states that it's "0 for explicit disconnect, non-zero for unexpected disconnect"  you're not _meant_ to be trying to decode it further.

denravonska:
> The documentation states that it's "0 for explicit disconnect, non-zero for unexpected disconnect" you're not _meant_ to be trying to decode it further.

In that case I'd argue that both the documentation and the implementation are wrong.

karlp:
The documentation matches the implementation.  Should the implementation be made more like other functions where rc _can_ be decoded with mosq_errstr()?  sure, that sounds perfectly reasonable, but that's an enhancement.

denravonska:
Depending on if it's a conscious decision or not you can view it as either a bug or a change request. Either way, it would be handy if the return codes are used consistently, especially when the function in question does use the provided define for positive outcomes.

ralight:
This is now fixed - I've done it in a slightly different and more comprehensive way than this PR, so I'm going to close this. Thanks for the request.

PR #444, opened 2019-12-06
add example of using trio as external loop

bkanuka:
This is an additional example of using an external event loop. This demonstrates using trio instead of the built-in loop or asyncio. This example closely mirrors loop_asyncio.py.

### DIFF
--- a/examples/loop_trio.py
+++ b/examples/loop_trio.py
@@ -1,0 +1,94 @@
+#!/usr/bin/env python3
+
+import socket
+import uuid
+import paho.mqtt.client as mqtt
+import trio
+
+client_id = 'paho-mqtt-python/issue72/' + str(uuid.uuid4())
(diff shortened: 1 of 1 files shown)


ralight:
Thanks very much!

examples/loop_trio.py
```diff
@@ -0,0 +1,94 @@
...
+        print("Socket opened")
+        self.sock = sock
+        self.sock.setsockopt(socket.SOL_SOCKET, socket.SO_SNDBUF, 2048)
+
+    def on_socket_register_write(self, client, userdata, sock):
+        print('large write request')
+        self._event_large_write.set()
+
+    def on_socket_unregister_write(self, client, userdata, sock):
+        print(f"finished large write")
```
hannesa2, line 50: @ralight This commit break the `master`

PR #466, opened 2020-03-02
Fix for sending MQTTv5 message properties when message is queued

yawor:
I've found that when I publish multiple messages at once with some properties attached, another client receives all the messages, but some of them arrive with properties stripped. After some digging I've found that this happens if I send more than 20 messages, which is a default value for max inflight messages.
This PR fixes that. I've also found another location in the code where properties were not published which I've also corrected in this PR.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -2834,7 +2834,8 @@ class Client(object):
                             m.payload,
                             m.qos,
                             m.retain,
-                            m.dup
+                            m.dup,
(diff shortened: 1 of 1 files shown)


yawor:
Is there any chance of merging this PR? Without it this library is seriously broken when MQTTv5 properties are used together with a high number of messages published from a client using it.
@PierreF?

PierreF:
Thank for your patience and the fix, which indeed make MQTTv5 hard to use.

PR #480, opened 2020-04-02
Added MQTTv5 RPC math client and server

fpagliughi:
Signed-off-by: fpagliughi <fpagliughi@mindspring.com>

I creates MQTTv5 examples for the Paho C++ and Rust libraries. These are for the Python library, and should be compatible with the examples in the other languages.  So the applications should be able to mix RPC client and server regardless of the implementation language.

### DIFF
--- a/examples/client_rpc_math.py
+++ b/examples/client_rpc_math.py
@@ -1,0 +1,110 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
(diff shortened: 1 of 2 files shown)


PierreF:
Thank for this example.

fpagliughi:
No problem. There are a few more that I want to port over as well, if I can ever find the time. There's a tiny "chat" app to show subscription "no local" and a new one to demonstrate topic aliases.

PR #485, opened 2020-04-11
Import ABC from collections.abc instead of collections for Python 3 compatibility

### LINKED ISSUE #450: Deprecation warning
While running a unit test on code using `paho-mqtt`, I've run into a deprecation warning:
```python
paho/mqtt/publish.py:127: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
  if not isinstance(msgs, collections.Iterable)
```

Since Python 3.8 has been released and 3.9 is next, I think this is worth fixing. I can write the PR later if the maintainers are interested in getting this fixed

tirkarthi:
Fixes #450

### DIFF
--- a/src/paho/mqtt/publish.py
+++ b/src/paho/mqtt/publish.py
@@ -21,6 +21,10 @@ broker, then disconnect and nothing else is required.
 from __future__ import absolute_import
 
 import collections
+try:
+    from collections.abc import Iterable
+except ImportError:
+    from collections import Iterable
 
 from . import client as paho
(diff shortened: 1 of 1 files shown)


PierreF:
Sorry for the delay to handle this PR. I've took PR from most recent to oldest and thus merged #497 which does the same.

tirkarthi:
No problem. Thanks @PierreF.

PR #486, opened 2020-04-22
Fix wrong variable assignments (they worked as comparisons instead). …

bgusach:
…Remove comparison to boolean value

Signed-off-by: Bor Gonzalez-Usach <bor.gonzalez.usach@contact-software.com>


PierreF:
Thanks for catching this !

PR #497, opened 2020-06-09
Fix for Python 3.9 moving Iterable to collections.abc

Brishen:
Signed-off-by: Brishen Hawkins <brishen.hawkins@gmail.com>

### DIFF
--- a/src/paho/mqtt/publish.py
+++ b/src/paho/mqtt/publish.py
@@ -21,6 +21,10 @@ broker, then disconnect and nothing else is required.
 from __future__ import absolute_import
 
 import collections
+try:
+    from collections.abc import Iterable
+except ImportError:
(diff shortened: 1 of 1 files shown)


Brishen:
The Travis-CI build is failing because the tests are outdated. See my pull request for the testing repo: https://github.com/eclipse/paho.mqtt.testing/pull/65

PierreF:
The Travis error seems unrelated to this PR. Sadly our tests fail from time to time, but I can't reproduce those failure making it hard to solve. Most of the time, re-running test "fix" them :(
On the Iterable, it indeed move to collections.abc since Python 3.3 (but it's probably only Python 3.9 which dropped it presence in collections).
Anyway, thank for your contribution !

PR #530, opened 2020-11-29
Add reconnect_on_protocol_failure argument

frederikaalund:
This is a non-breaking change.

I added the argument `reconnect_on_protocol_failure` to the `Client.__init__` function. It defaults to `True`, which aligns with the current semantics. When set to `False`, paho won't try to reconnect inside `_handle_connack` (it simply returns `MQTT_ERR_PROTOCOL` instead).

---

This feature request came from a bug report in sbtinstruments/asyncio-mqtt#26. The implicit reconnect inside `handle_connack` blocks the current thread. When we use the external event loop API (as we do in asyncio-mqtt), it blocks the event loop itself. That is very bad.

We will set `reconnect_on_protocol_failure=False` in asyncio-mqtt to avoid blocking the event loop. I'd recommend all users of the external event loop API to do the same.

I'm open to renaming stuff and moving the argument to another function if you think that aligns better with the current paho API design. Let me know what you think. :+1:

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -513,7 +513,7 @@ class Client(object):
     """
 
     def __init__(self, client_id="", clean_session=None, userdata=None,
-                 protocol=MQTTv311, transport="tcp"):
+                 protocol=MQTTv311, transport="tcp", reconnect_on_failure=True):
         """client_id is the unique client id string used when connecting to the
         broker. If client_id is zero length or None, then the behaviour is
         defined by which protocol version is in use. If using MQTT v3.1.1, then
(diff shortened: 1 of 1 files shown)


ralight:
I think this might be better as a more generic parameter like `reconnect_on_failure` and to cover other failures as well, would you mind changing to do that? If you fancied adding code to check more of the failure modes at the same time that would be great :)

frederikaalund:
Thanks for the review. :+1: 

> I think this might be better as a more generic parameter like reconnect_on_failure and to cover other failures as well, would you mind changing to do that? If you fancied adding code to check more of the failure modes at the same time that would be great :)

Good point. I generalized it to `reconnect_on_failure`. As far as I can tell, there is only one other place where this changes things: inside `loop_forever`.

frederikaalund:
Any updates on this? :) I'd like to have this as part of asyncio-mqtt before I release 1.0.0.

ralight:
I merged this to master by mistake, but have removed it and merged to 1.6.x instead. I've also added a test that exercises this feature. Thanks very much for the PR, and sorry for taking so long to get to it!

PR #547, opened 2021-01-10
Excluded f-strings files from flake linting.

lucacillario:
# New features and improvements
None

# Breaking changes
None

# Bug fixes
- Excluded f-strings files from flake linting

# Related issue
None

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,16 +2,20 @@
 envlist = py{27,35,36,37,38}
 
 [testenv:py27]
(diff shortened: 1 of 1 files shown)


lucacillario:
Tests are bit flaky. I suspect it is due to too many concurrent jobs in the CI pipeline. I would try to set the number of concurrent jobs to 3 in travis' settings (unfortunately I can't do it), otherwise i could divide them into two different stages.

ralight:
I don't believe this is necessary any more, the example that had the f-string has been updated to remove them. Thanks for taking the time to look into it this way!

PR #556, opened 2021-03-15
Travis CI: Remove unnecessary Python f-strings

cclauss:
These Python f-strings are not necessary and are breaking Travis CI jobs.
https://travis-ci.org/github/eclipse/paho.mqtt.python/pull_requests


ralight:
Thanks very much for this.

Issue #377, opened 2019-04-24
Send queue has a size limit of 65555 when using QoS > 0


yschroeder:
Contrary to the documentation, which states that there is an unlimited queue size by default, the send queue for messages with QoS > 0 is limited to 65555 entries via this `if` statement:
https://github.com/eclipse/paho.mqtt.python/blob/master/src/paho/mqtt/client.py#L1210

I think the code could be changed so that message ids are assigned later, when the message should actually be sent and a message id is available again.

fargiolas:
Any insight from the devs about this issue? a brief look at the code suggests this is true and the queue is actually limited to 65535 messages given it's not an actual queue but a dict indexed by message ids. It would be great to have some confirmation about this, I wanted to use paho as my sole message queue but with this constraint it seems I'd start losing data after about 18 hours (sending 1 sample per second) of network down.

yschroeder:
I will shamelessly summon @ralight, as I think this issue got lost between the support requests from Python beginners.

As my above link is outdated, I will describe the problem again:

[Here](https://github.com/eclipse/paho.mqtt.python/blob/225ab3757f6818ba85eb80564948d1c787190cba/src/paho/mqtt/client.py#L1269) the message ID for an outgoing message is generated via `_mid_generate()`. It assigns the numbers from 1 to 65535 as message IDs in a round-robin style.

[Here](https://github.com/eclipse/paho.mqtt.python/blob/225ab3757f6818ba85eb80564948d1c787190cba/src/paho/mqtt/client.py#L1291) the generated message ID is checked if it already exists in the dictionary that holds outgoing messages. If the message ID is already in there, the message is discarded. This might happen when the network connection is gone while 2^16 messages are requested to be published. In fact, this dictionary should be a queue and allow arbitrary numbers of messages.

[Here](https://github.com/eclipse/paho.mqtt.python/blob/225ab3757f6818ba85eb80564948d1c787190cba/src/paho/mqtt/client.py#L1295) the new message is put into the dictionary if the message ID does not exist there (the previous message with the same ID is already sent successfully).

In my opinion, this behavior is a bug when using QoS 1 or QoS 2. I expect all my messages that I put into the MQTT stack to reach the broker eventually.

fargiolas:
I believe this comes from the protocol limiting IDs to 16bits, but as far as I can tell nothing prevents an implementation to store messages in an independent local queue and assign them an ID as soon as they become available for reuse.

[...]

PR #568, opened 2021-05-01
Updated documentation: max message queue size

JsBergbau:
I've also stumbled upon the max message queue size of 65555, see https://github.com/eclipse/paho.mqtt.python/issues/377
So unless this is fixed documentation should tell about that fact. Knowing this can save a lot of time, hassles and headache.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -249,7 +249,7 @@ max_queued_messages_set()
 
 Set the maximum number of outgoing messages with QoS>0 that can be pending in the outgoing message queue.
 
-Defaults to 0. 0 means unlimited. When the queue is full, any further outgoing messages would be dropped.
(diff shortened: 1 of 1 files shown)


yschroeder:
Maybe you want to clarify that this is indeed 65555 and not 65535 with a typo?

Something like:

`Defaults to 0. 0 means unlimited, but due to implementation currently limited to 65555 (65535 messages in queue + 20 messages in flight). When the queue is full, any further outgoing messages would be dropped.`

JsBergbau:
Thanks for the hint. Documentation is updated. Hopefully this pull request is merged soon.

JsBergbau:
This is just a minor documentation fix, however it is still not merged. What is else required that it can be merged?

ralight:
Thank you, and sorry for not getting to this earlier.

PR #569, opened 2021-05-05
Close the connection if _send_pingreq raises an exception

BertKleewein:
Signed-off-by: Bert Kleewein <bertk@microsoft.com>


ralight:
Thank you, I think this makes complete sense.

PR #572, opened 2021-05-15
Decorators to register client callback functions

MhhhxX:
I implemented decorators to register the callback functions in the Client class. This allows to write code like this:

```
from paho.mqtt.client import Client

host = "localhost"
client = Client()
topic = "MyTopic/SubTopic"

@client.connect_callback()
    # handle the callback
    pass


@client.topic_callback(topic)
def on_subtopic(client, userdata, message):
    # handle incoming message from MyTopic/SubTopic feed
    msg = message.payload.decode()

if __name__ == '__main__':
    client.connect(host)
    client.subscribe(topic)
    client.loop_forever()
```
I think this would be a nice addition to this library. In my opinion it makes the callback handling stuff more comfortable to use. What do you think?

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -1857,6 +1857,12 @@ class Client(object):
         """
         self._on_log = func
 
+    def log_callback(self):
+        def decorator(func):
+            self.on_log = func
+            return func
+        return decorator
+
     @property
     def on_connect(self):
         """If implemented, called when the broker responds to our connection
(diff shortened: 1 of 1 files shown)


ralight:
I like this, thank you. I manually merged it on to the 1.6.x branch without updating github first and now github doesn't recognise that this should be closed. The commit is here https://github.com/eclipse/paho.mqtt.python/commit/d8aa1c18523af24e13533751c64fbda89e35a286 and the next commit includes a test for most of the decorators.

PR #578, opened 2021-06-21
Fix wrong usage of socket timeout

gilbertwong96:
The correct usage for `socket.create_connection` is below:

>socket.create_connection(address[, timeout[, source_address]])
>Connect to a TCP service listening on the Internet address (a 2-tuple (host, port)), and return the socket object. This is a higher-level function than socket.connect(): if host is a non-numeric hostname, it will try to resolve it for both AF_INET and AF_INET6, and then try to connect to all possible addresses in turn until a connection succeeds. This makes it easy to write clients that are compatible to both IPv4 and IPv6.
>Passing the optional timeout parameter will set the timeout on the socket instance before attempting to connect. If no timeout is supplied, the global default timeout setting returned by getdefaulttimeout() is used.
>If supplied, source_address must be a 2-tuple (host, port) for the socket to bind to as its source address before connecting. If host or port are ‘’ or 0 respectively the OS default behavior will be used.
>Changed in version 3.2: source_address was added.

Also the semantics of `connect timeout` and `keepalive timeout` are totally different.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -568,6 +568,7 @@ class Client(object):
         self._sockpairR, self._sockpairW = (None, None,)
         self._sockpairR, self._sockpairW = _socketpair_compat()
         self._keepalive = 60
+        self._connect_timeout = 5.0
         self._message_retry = 20
         self._last_retry_check = 0
         self._client_mode = MQTT_CLIENT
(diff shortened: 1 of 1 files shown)


ralight:
I agree, the old behaviour is incorrect. Thanks for spotting this and submitting the PR.

PR #587, opened 2021-07-21
Fix None comparation as PEP

hildogjr:
This fix is the same as Fix #584 but trying to achieve the right eclipsefdn/eca e-mail configurations.


ralight (approved):
This all looks fine, thank you.

PR #589, opened 2021-08-03
loop: Use original loop_stop API to stop loop of loop_forever

shianehao:
In some case needs to stop loop_forever loop, checked code of  it found had a internal terminal flag that only can set on loop_stop. Re-order loop_stop's code to handle this requested.


ralight:
I don't think overloading this function is the correct thing to do. Calling `disconnect()` will exit the loop, is that not sufficient for you?

shianehao:
@ralight 

Yes, after checked with disconnect() also work as my expected.

PR #593, opened 2021-08-11
Fix identation and None comparation

hildogjr:


src/paho/mqtt/reasoncodes.py
```diff
@@ -120,7 +120,7 @@ def __init__(self, packetType, aName="Success", identifier=-1):
         }
         if identifier == -1:
             if packetType == PacketTypes.DISCONNECT and aName == "Success":
-                  aName = "Normal disconnection"
+                    aName = "Normal disconnection"
```
ralight, line 123: This makes the indentation 8 spaces when it should surely be 4 spaces.

ralight:
Thanks very much!

PR #616, opened 2021-11-03
Add v1.6.0 connection timeout change to change log

ncoghlan:
I was digging into timeout management to try to figure out a problem we've been seeing, and noticed that this had changed, but wasn't in the change log.


PierreF:
Thank for your contribution

PR #617, opened 2021-11-03
Set client_id as thread name

mosauter:
I had a couple of problems in my code where I pulled up a couple of connections/clients and randomly it happened that in the loop an exception occurred. When it happens the exception always starts like this before the stacktrace:

```
Exception in thread Tread-$some_number:
```

... which gives not too much info where the exception happened, as the stack trace ends always at `loop_forever` and the interesting part would most of the time be who actually started the thread that caused the exception.

The change gives control over how the thread that is created by `loop_start` is named. To not change any API at all I opted for the decoded client id. If not explicitly set, there will be the usual "broker-generated string" as name, but for me that does not give any less information than the mentioned above `Thread-$some_number`. But if you explicitly set the `client_id` (which needs to be unique, so no name clashes possible as far as I can see) the thread will reflect that change, so in the exception it would print:

```
Exception in thread manually_set_client_id:
```

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -1808,7 +1808,7 @@ class Client(object):
 
         self._sockpairR, self._sockpairW = _socketpair_compat()
         self._thread_terminate = False
-        self._thread = threading.Thread(target=self._thread_main)
+        self._thread = threading.Thread(target=self._thread_main, name=f"paho-mqtt-client-{self._client_id.decode()}")
         self._thread.daemon = True
         self._thread.start()
 


src/paho/mqtt/client.py
```diff
@@ -1794,7 +1794,7 @@ def loop_start(self):
 
         self._sockpairR, self._sockpairW = _socketpair_compat()
         self._thread_terminate = False
-        self._thread = threading.Thread(target=self._thread_main)
+        self._thread = threading.Thread(target=self._thread_main, name=self._client_id.decode())
```
PierreF, line 1797:
```suggestion
        self._thread = threading.Thread(target=self._thread_main, name=f"paho-mqtt-client-{self._client_id.decode()}")
```

I think the thread name should include that it's coming from paho. The client_id may not include the fact that it's a MQTT client and the program may not only be a MQTT client.

PierreF (commented):
Thank for your contribution and your patience. I've made a small suggestion change.

Also could you add a Signed-Off to your commit ? This is a requirement from Eclipse to acknowledge you agree with the ECA.

mosauter:
Thanks for the review. 

I've applied your requested change and added the sign-off.

PR #622, opened 2021-11-18
Use time_func for timeout in wait_for_publish

jurijsteblaj:
Hello,

MQTTMessageInfo.wait_for_publish uses time.time, which returns non-monotonic clock time. It should probably be replaced with time_func defined in the same file, which uses time.monotonic when available.

Signed-off-by: Jurij Šteblaj <jurij@steblaj.si>


thijstriemstra (approved):

PierreF:
Thank for your contribution.

PR #629, opened 2021-12-08
Remove broken conversion to int in ReasonCodes.unpack on Python2

### LINKED ISSUE #628: Error unpacking ReasonCodes in Python2.7
Hello there, 

I've got a test that used to run fine with paho-mqtt 1.5.1 on Python 2.7 but now it's failing. Here's the traceback:

```
  [irrelevant frames]
   client.loop()
  File "c:\...\paho_mqtt-1.6.1-py2.7.egg\paho\mqtt\client.py", line 1120, in loop
    return self._loop(timeout)
  File "c:\...\paho_mqtt-1.6.1-py2.7.egg\paho\mqtt\client.py", line 1164, in _loop
    rc = self.loop_read()
  File "c:\...\paho_mqtt-1.6.1-py2.7.egg\paho\mqtt\client.py", line 1556, in loop_read
    rc = self._packet_read()
  File "c:\...\paho_mqtt-1.6.1-py2.7.egg\paho\mqtt\client.py", line 2439, in _packet_read
    rc = self._packet_handle()
  File "c:\...\paho_mqtt-1.6.1-py2.7.egg\paho\mqtt\client.py", line 3045, in _packet_handle
    return self._handle_disconnect()
  File "c:\...\eggs\paho_mqtt-1.6.1-py2.7.egg\paho\mqtt\client.py", line 3220, in _handle_disconnect
    reasonCode.unpack(self._in_packet['packet'])
  File "c:\...\paho_mqtt-1.6.1-py2.7.egg\paho\mqtt\reasoncodes.py", line 166, in unpack
    c = ord(c)
TypeError: ord() expected string of length 1, but int found
```

Here's the body of `unpack`:

```
def unpack(self, buffer):
    c = buffer[0]
    if sys.version_info[0] < 3:
        c = ord(c)
    name = self.__getName__(self.packetType, c)
    self.value = self.getId(name)
    return 1
```

Apparently the reason for this problem is that the argument `buffer` used to be a `bytes` object (or `str` in Python2) on version 1.5.1 but now it's a `bytearray`... which causes `buffer[0]` to return an `int` instead of a `str`/`bytes` object with length of 1... and then subsequently the call to `ord` fails. [Here where the type was defined on 1.5.1](https://github.com/eclipse/paho.mqtt.python/blob/v1.5.1/src/paho/mqtt/client.py#L593), and [here on 1.6.1](https://github.com/eclipse/paho.mqtt.python/blob/v1.6.1/src/paho/mqtt/client.py#L570).

The solution seems clear to me: remove the `if sys.version_info[0] < 3:` branch, as there is no need to check if python2 or 3 if we've got already an int.

bgusach:
Fixes #628

### DIFF
--- a/src/paho/mqtt/reasoncodes.py
+++ b/src/paho/mqtt/reasoncodes.py
@@ -162,8 +162,6 @@ class ReasonCodes:
 
     def unpack(self, buffer):
         c = buffer[0]
-        if sys.version_info[0] < 3:
-            c = ord(c)
         name = self.__getName__(self.packetType, c)
         self.value = self.getId(name)
         return 1


PierreF:
Thank for your contribution. This PR is no longer needed following drop of Python 2 support done in #767

Issue #653, opened 2022-03-22
DeprecationWarning: ssl.PROTOCOL_TLS is deprecated


maranov:
Per [Python ssl docs](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS), the `PROTOCOL_TLS` has been deprecated, replaced with `PROTOCOL_TLS_CLIENT/SERVER`. This setting is used in the client at:
<https://github.com/eclipse/paho.mqtt.python/blob/9782ab81fe7ee3a05e74c7f3e1d03d5611ea4be4/src/paho/mqtt/client.py#L791>

From what I understood, the `PROTOCOL_TLS_CLIENT` should be the correct replacement (for client connections), but it has some side effects:
> The protocol enables [CERT_REQUIRED](https://docs.python.org/3/library/ssl.html#ssl.CERT_REQUIRED) and [check_hostname](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.check_hostname) by default.

Looking at the client code, this seems to be the defaults for `tls_set` as well. Replacing the `PROTOCOL_TLS` with `PROTOCOL_TLS_CLIENT`, while using `PROTOCOL_TLS` as a fallback for Python 2.7 should remove the deprecation warning.

maranov:
I could open a PR if the proposed solution sounds like a good idea.

timhaines:
@maranov sounds good to me.  Would love to see this warning disappear. 👍

MattBrittan:
As PR #665 has now been merged I believe this issue can be closed. Note that there are a few other related issues that will remain open (the documentation does not really match the code).

Note: This is part of an exercise to clean up old issues so that the project can move forwards. Due to the number of issues being worked through mistakes will be made; please feel free to reopen this issue (or comment) if you believe it's been closed in error.

poonkothaip:
Hi
Using python 3.12 and using below code to connect -

def create_ftps_client(server_hostname, server_port, certfile, keyfile):
  
    context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
    context.load_cert_chain(certfile=certfile, password=keyfile)
    ftps = FTP_TLS(context=context)
    try:

        ftps.connect(server_hostname, server_port)
        ftps.auth()
        print("FTPS connection established successfully")
        return ftps
    except ssl.SSLError as e:
        print(f"Failed to connect to the server: {e}")
        return None


[...]

PR #665, opened 2022-06-10
Fix Python >=3.10 PROTOCOL_TLS DeprecationWarning

maranov:
PROTOCOL_TLS_CLIENT/_SERVER are attempted to be selected before falling back to the legacy PROTOCOL_TLS setting. The side effects of using PROTOCOL_TLS_CLIENT is that CERT_REQUIRED and check_hostname are enabled by default, but these were explicitly used as a default already and remain to be the defaults even if PROTOCOL_TLS_CLIENT is unavailable.

The settings in `examples\client_pub_opts.py` and `examples\client_sub_opts.py` are unaffected.

Bug: https://github.com/eclipse/paho.mqtt.python/issues/653

The warning is now gone:
image

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -286,7 +286,7 @@ tls_set()
 .. code:: python
 
     tls_set(ca_certs=None, certfile=None, keyfile=None, cert_reqs=ssl.CERT_REQUIRED,
-        tls_version=ssl.PROTOCOL_TLS, ciphers=None)
+        tls_version=ssl.PROTOCOL_TLS_CLIENT, ciphers=None)
 
 Configure network encryption and authentication options. Enables SSL/TLS support.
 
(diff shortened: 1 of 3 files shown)


sytherisop:
> 

dude i am facing this error u know how to fix this

zhanglongqi:
May I know why this is not merged since it has been existed for several months.

henloef:
> May I know why this is not merged since it has been existed for several months.

It is not fixed because no-one is actively maintaining this package, according to the mailing list https://www.eclipse.org/lists/paho-dev/msg04742.html

Furthermore, pypi lists python 3.10 as supported, while it is not actually tested according to the tox specification [/workflows/tox.yml](https://github.com/eclipse/paho.mqtt.python/blob/a4cb435ca2864d073ea3e0e18b0407e4bbe85b16/.github/workflows/tox.yml#L9), which might be the reason why this bug is allowed to exists

PierreF:
Thanks for this contribution. To my understanding it has one minor change: hostname will be checked when the previously wasn't. This seems a good change, the next release will be with breaking change so I'll update the Changelog to include this "breaking change".

For the ECA check, I believe the automated check is wrong, it probably get confused because the Author & Committer email aren't the same. Anyway you author & signed-off email had a valid ECA signature which is fine for me.

PR #674, opened 2022-08-14
Upgrade ConnectionError to OSError

enkiusz:
Upgrade ConnectionError to OSError to catch more TCP failures (for example timeouts). Minor exception handling cleanups.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -2370,7 +2370,7 @@ class Client(object):
                 command = self._sock_recv(1)
             except BlockingIOError:
                 return MQTT_ERR_AGAIN
-            except ConnectionError as err:
+            except OSError as err:
                 self._easy_log(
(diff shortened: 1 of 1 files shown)


gonzalocasas:
+1 to this because `ConnectionError` is causing the client to be unusable in Python 2.7.

PierreF:
Hello and thanks for your patience.

I'm not sure what issue we want to solve. It seems we should never get any TimeoutError since the socket is set to setblocking(0). So I guess we receive BlockingIOError instead of a possible TimeoutError ?

In case of Timeout, shouldn't we prefer the handling done in BlockingIOError: retry later rather that disconnect ?

PierreF:
Closing this PR, since #797 which fixed another issue (SSL error not being catched) seems to result in very similar fixed.

If there is some case not covered by #797 feel free to re-open the PR and explaining what issue you solve this your PR.

Issue #505, opened 2020-08-06
AttributeError: 'NoneType' object has no attribute 'recv'


freespy:
```
ERROR mqtt error 'NoneType' object has no attribute 'recv' 2020-08-06 23:12:47
ERROR Traceback (most recent call last):
  File "/data/dtuapi/dtuapi.py", line 117, in run
    self.client.loop_forever()
  File "/usr/local/lib/python2.7/site-packages/paho/mqtt/client.py", line 1787, in loop_forever
    rc = self.loop(timeout, max_packets)
  File "/usr/local/lib/python2.7/site-packages/paho/mqtt/client.py", line 1182, in loop
    rc = self.loop_read(max_packets)
  File "/usr/local/lib/python2.7/site-packages/paho/mqtt/client.py", line 1573, in loop_read
    rc = self._packet_read()
  File "/usr/local/lib/python2.7/site-packages/paho/mqtt/client.py", line 2276, in _packet_read
    byte = self._sock_recv(1)
  File "/usr/local/lib/python2.7/site-packages/paho/mqtt/client.py", line 660, in _sock_recv
    return self._sock.recv(bufsize)
AttributeError: 'NoneType' object has no attribute 'recv'
```

valebes:
I have the same error.

BadWolf42:
Hello,

I'm facing the same issue and could reproduce quite easily/often.
Enclosed code : [paho-mqtt-issue505.py.txt](https://github.com/eclipse/paho.mqtt.python/files/6542440/paho-mqtt-issue505.py.txt).
Associated log files : [20210225-2310.log](https://github.com/eclipse/paho.mqtt.python/files/6542443/20210225-2310.log), [20210225-2315.log](https://github.com/eclipse/paho.mqtt.python/files/6542445/20210225-2315.log), [20210225-2317.log](https://github.com/eclipse/paho.mqtt.python/files/6542447/20210225-2317.log)

Using `Python 3.7.3 (default, Jan 22 2021, 20:04:44)` and `paho-mqtt version 1.5.1` on 
`Linux mydevhost 4.19.0-16-amd64 #1 SMP Debian 4.19.181-1 (2021-03-19) x86_64 GNU/Linux`

I think this issue is coming from the fact that the MainThread destroys the object, which calls `_reset_sockets()`, then `_sock_close()` and sets `self._sock` to None...

But at the same time, `self._sock` is taken for granted here and read by a worker Thread:
[...]

PR #686, opened 2022-11-17
Handle edge case where socket is None

RobertDeRose:
* The following exception happens randomly on startup:

```
Exception has occurred: AttributeError
'NoneType' object has no attribute 'recv'
  File "paho/mqtt/client.py", line 640, in _sock_recv
```

* The root cause is unknown, but this allows the library to reconnect gracefully.

Fixes https://github.com/eclipse/paho.mqtt.python/issues/505

### DIFF
--- a/src/paho/mqtt/__init__.py
+++ b/src/paho/mqtt/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "1.6.2"
+__version__ = "1.6.1"
 
 
 class MQTTException(Exception):
--- a/src/paho/mqtt/__init__.py
+++ b/src/paho/mqtt/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "1.6.1"
+__version__ = "1.6.2"
 
 
 class MQTTException(Exception):
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -643,6 +643,10 @@ class Client(object):
         except ssl.SSLWantWriteError:
             self._call_socket_register_write()
             raise BlockingIOError
+        except AttributeError as err:
+            self._easy_log(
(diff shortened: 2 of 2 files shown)


RobertDeRose:
@ralight I've search the issues and see others have reports this issue and stated that calling `loop_start` instead of `loop_forver`, however, `loop_start` just calls `loop_forever` on its own and this issue can happen and its not possible to catch it as it's inside an unmanaged thread.

This might not be to the correct fix, but it fixes it for me and I would assume others as if the Socket is None, there is no connection and the reconnect logic seems to kick in

RobertDeRose:
@ralight and @PierreF hate to be a bother here, but this is causing me some headaches in a production environment and while I could make a fork and build my own version, It would be nice if one of the maintainers could confirm that this appears to be the correct fix. Thanks!

gabrielgbs97:
Would like to see this one on main, had same issue... Still couldn't face the root of cause

RobertDeRose:
@gabrielgbs97 

I gave up getting this PR merged.  I ended up doing my own thread management

Just start your own thread, call `loop_forver` inside a `try..except` block and recall `loop_fover` if an exception is thrown
The thread is simple a `while True:` around the `try...except`

src/paho/mqtt/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "1.6.1"
+__version__ = "1.6.2"
```
PierreF, line 1: We should not change the version in a PR

PierreF:
Thank for your contribution.

As I said in #505, I think the real fix would be to audit concurrency access and improve either locking or improve documentation on what could be called concurrently. In the meantime your fix seems simple and correct the the described problem.

PR #695, opened 2022-12-15
Adding User Data Getter

arichadda:
Adding a getter method for `userdata` for the case of accessing the data outside of a callback. I found this useful when writing unit tests for publish/subscribe events.

Please let me know if there's a reason a getter method was not included in the original codebase. Thanks for supporting this project!

Signed-off-by: arichadda <ari.chadda@gmail.com>


PierreF:
Thank for your contribution.

I think we didn't had any accessor for userdata because you could always pass it with another way (e.g. pass a tuple (client, userdata) instead of just client). But I agree it could make some usage easier and don't incur additional maintenance work.

PR #698, opened 2022-12-30
Respoect max_inlight message count in reconnect

### LINKED ISSUE #492: At reconnect, the client doesn't seem to take max_inflight_messages into account
Hi,
We've been using your service for some time now and are very happy about the results. Thanks for this! 

Our setup is the following:
Sensor - (Dash 7)-> gateway -(mqtt)-> server
On the server, we do authentication of the gateway but also the sensor sending the information. When an unauthenticated sensor sends a message through our gateway, the gateway gets disconnected from the broker as per the following documentation: http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc39871804 . 

Now I've created some code to manually delete the last entry and this seems to work but when the service reconnects, it seems to publish all messages in the _out_messages queue while we've set the max_inflight_messages to 1. I've looked at the code and looks like there's no check for this (https://github.com/eclipse/paho.mqtt.python/blob/master/src/paho/mqtt/client.py#L3054). Shouldn't we check here if we're not publishing to many messages at once?

Thank you in advance!

Kind regards,
Liam

Juha-Ylikoski-Treon:
Previously if the message queue filled up for some reason like bad connection and the client disconnected (due to e.g. bad connection), it would send all of it's messages in _out_messages-queue and not respect _max_inflight_messages parameter. With some brokers and when we had 1000s of messages in queue and all being sent with qos 1 without any regard to the _max_inflight_messages, the broker would disconnect us or the connection would again break down.

This patch makes sure _max_inflight_messages is respected when reconnect happens for any reason.

I believe this also fixes #492

Signed-off-by: Juha Ylikoski <juha.ylikoski@treon.fi>


PierreF:
Thanks for your contribution. I've written some test in https://github.com/eclipse/paho.mqtt.python/commit/a2574e61aea8ebb3bb5df3926f4b8771ca0afe8e (pushed on master) which show the problem exist on master.

It also show that there is another problem: message are published twice, once with mid=original_mid + N and once with mid=original_mid & dup = 1.

This issue still exists on this PR. I haven't yet looked whether the two issues are unrelated (and then the PR should be merged as-is) or they are related and should be fixed together.

Issue #706, opened 2023-02-27
pip install deprecation warning


andypiper:
Hello from a member of the original Paho team 🙂 thanks for continuing the great work!

On macOS Ventura, Python 3.11, pip 23.0

```
 DEPRECATION: paho-mqtt is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559
  Running setup.py install for paho-mqtt ... done
```

There are workarounds listed (and the installation works and the library is fine), but Paho probably needs to be updated for the new `pyproject.toml` library metadata format.

semuadmin:
Great library, but same issue here. Any timeline on pyproject.toml support?

MattBrittan:
I believe PR #774 (now committed) addresses this; would appreciate it if someone could confirm.

pswsm:
Hello @MattBrittan , indeed this issue has been addressed, so it can be closed.

cc: @PierreF

PR #729, opened 2023-05-31
Migrate pyproject toml

semuadmin:
setup.py is now [deprecated](https://pip.pypa.io/en/stable/reference/build-system/setup-py/) and is scheduled to be dropped altogether in pip 23.

This PR makes a number of changes to update the project build framework:

1. It updates the paho.mqtt project structure to use pyproject.toml, while retaining setuptools as a back-end build platform.
2. It uses pyproject.toml tool configuration sections were available.
3. It drops support for Python versions <=3.6 as these are now end of life.
4. It adds a minimal VS Code workflow including the following tasks (Eclipse IDE users can simply ignore these):
 - The Build task will produce wheel (*.whl) and sdist (*.tar.gz) install packages in the dist folder which can be uploaded to PyPi using twine.
 - The Install Locally task will install this wheel using pip.
 - Other VS Code tasks have been added mirroring the current GitHub Actions (GHA) CI worfklows.

NB: this PR only changes the build setup - no changes have been made to existing application code or test cases, so some existing CI workflows may still fail if they were failing previously.

Fixes https://github.com/eclipse/paho.mqtt.python/issues/706


README.rst
```diff
@@ -5,7 +5,7 @@ This document describes the source code for the `Eclipse Paho <http://eclipse.or
 
 This code provides a client class which enable applications to connect to an `MQTT <http://mqtt.org/>`_ broker to publish messages, and to subscribe to topics and receive published messages. It also provides some helper functions to make publishing one off messages to an MQTT server very straightforward.
 
-It supports Python 2.7.9+ or 3.6+.
+It supports Python 3.7+.
```
cclauss, line 8: OPTIONAL: Python EOLs in < 30 days so support for it should probably be dropped as well but that can be done in a separate pull request in the coming weeks.  https://devguide.python.org/versions

Vagrantfile
```diff
@@ -1,17 +1,11 @@
 Vagrant.configure("2") do |config|
   # The base OS 
-  config.vm.box = "ubuntu/trusty64"
+  config.vm.box = "ubuntu/bionic64"
```
cclauss, line 3: jammy?  https://wiki.ubuntu.com/Releases

README.rst
semuadmin: > OPTIONAL: Python EOLs in < 30 days so support for it should probably be dropped as well but that can be done in a separate pull request in the coming weeks. https://devguide.python.org/versions

Agreed - I'm in the process of dropping 3.7 from my repos as well

Vagrantfile
semuadmin: > jammy? https://wiki.ubuntu.com/Releases

TBH not sure what you want to do with Vagrant here in terms of legacy support - clearly there are a couple of LTS Ubuntu releases since bionic. Personally I'd go for 22.04.2 LTS (Jammy Jellyfish).

semuadmin: have updated to jammy

semuadmin:
@cclauss What's the game plan with regard to the lint_python pytest failures in GHA? AFAICS These are nothing to do with anything I've changed as part of this PR and I'm not in a position to resolve them?

semuadmin:
Is this repo still actively maintained?

redundant/tox.ini
PierreF, line 1: Without tox, how do we run linters ? I want to kept the ability to run linters & tests locally and not only relying on Github actions.

With tox, I only need to run two commands: `tox -e py` and `tox -e lint`. Everything was handled:
* creation of virtualenv, installation of dependency
* running test and linters command (for linters it's valuable, since we have multiple linters)

tox might not be the best tool for this usecase (here we don't use tox for it's ability to run on multiple Python versions, I use different Docker containers for that) that but it works and it's a well-known tools. I think we should kept it.

PierreF (changes requested):
Thanks for this contribution. Could you add a Signed-Off to your commit ? This is a requirement from Eclipse to acknowledge you agree with the ECA.

redundant/tox.ini
akx: > Without tox, how do we run linters ?

One common option is https://pre-commit.com/ which can also be [easily](https://github.com/python-babel/babel/blob/2a1709a7768f6f07c3d2dbfdb03d3c8a6bd80aef/.github/workflows/ci.yml#L16-L22) run in CI.

cclauss: It would probably be best to upgrade to **pre-commit** to automate the fast tests like ruff, codespell, black.
* https://pre-commit.com

semuadmin:
sorry no longer have the bandiwidth to progress this

cclauss:
@akx Given that both @semuadmin and I closed our PRs, perhaps you can jump in with your automated tool for creating these PRs.

akx:
@cclauss I don't have an automated tool per se - `pyproject-migrator` still needs hand-holding...

PR #753, opened 2023-10-03
application controlled acknowledgements to match Java behaviour

### LINKED ISSUE #348: Prevent ack of received message until final

petersilva:
The existing python library acknowledges all messages received after calling the on_message() callback configured by the application.  This PR adds manual_acks=False as an argument to client.py/__init__() function.  When manual_acks=True, the application must acknowledge receipt of every message using the newly created ack(mid,qos) entry point.
  
* Closes #348 

This PR addresses issue  #348 to allow applications to delay acknowledgement until it has completely received it. There was a first PR (#554) submitted two years ago. and got some great feeback from it:

* wanted QoS==2 support (the PR addressed only QoS==1)
* worry about reception order (is the library responsible for ensuring that acknowledgements are sent in the order the messages were received.)

After much too and fro, it was fairly clear that the library cannot be responsible for ordering. Then it occurred to me, that the Java implementation exists, and the equivalent functionality has been present for a number of years, and so chances are that it is doing "the right thing" tm.

So this PR uses the java implementation as a model, and slavishly implements the same thing in Python.

To illustrate the relationship to Java implementation, we can compare some tidbits:

So in python we have:

```

in _handle_publish(self):
     .
     .
     .
      elif message.qos == 1:
            self._handle_on_message(message)
            if self._manual_ack:
                return MQTT_ERR_SUCCESS
            else:
                return self._send_puback(message.mid)

```

vs. Java:

```

in:     private void handleMessage(MqttPublish publishMessage)

                  // If we are not in manual ACK mode:
                if (!this.manualAcks && publishMessage.getMessage().getQos() == 1) {
                        this.clientComms.internalSend(new MqttPubAck(MqttReturnCode.RETURN_CODE_SUCCESS,
                                        publishMessage.getMessageId(), new MqttProperties()),
                                        new MqttToken(clientComms.getClient().getClientId()));
                }

```

There is a need for an implementation of a call to be used by the library user to actually send the acknowledgements.  we can compare code here also:

python:

```
    def ack( self, mid, qos ):
        """
           send an acknowledgement for a given message id. (stored in message.mid )
           only useful in QoS=1 and auto_ack=False
        """
        if self._manual_ack :
            if qos == 1:
                return self._send_puback(mid)
            elif qos == 2:
                return self._send_pubcomp(mid)

        return MQTT_ERR_SUCCESS

```

versus Java:

```

       public void messageArrivedComplete(int messageId, int qos) throws MqttException {
                if (qos == 1) {
                        this.clientComms.internalSend(
                                        new MqttPubAck(MqttReturnCode.RETURN_CODE_SUCCESS, messageId, new MqttProperties()),
                                        new MqttToken(clientComms.getClient().getClientId()));
                } else if (qos == 2) {
                        this.clientComms.deliveryComplete(messageId);
                        MqttPubComp pubComp = new MqttPubComp(MqttReturnCode.RETURN_CODE_SUCCESS, messageId, new MqttProperties());
                        // @TRACE 723=Creating MqttPubComp due to manual ACK: {0}
                        log.info(CLASS_NAME, "messageArrivedComplete", "723", new Object[] { pubComp.toString() });

                        this.clientComms.internalSend(pubComp, new MqttToken(clientComms.getClient().getClientId()));
                }
        }

```

In the java, there was an existing messageArrivedCompleted entry point to adjust.  In python, nothing similar was obvious, so modified _handrel which would only apply in QoS==2... to suppress sending of pubcomp when manual_acks is active.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -197,6 +197,12 @@ transport
     set to "websockets" to send MQTT over WebSockets. Leave at the default of
     "tcp" to use raw TCP.
 
+manual_ack
+    defaults to False, allowing the library to acknowledge messages before
+    passing them to on_message callback.  When set to True, every message
+    must be manually acknowledged by application call to 
+    client.ack( *message.mid* )
+
 
 Constructor Example
 ...................
--- a/README.rst
+++ b/README.rst
@@ -201,7 +201,7 @@ manual_ack
     defaults to False, allowing the library to acknowledge messages before
     passing them to on_message callback.  When set to True, every message
     must be manually acknowledged by application call to 
-    client.ack( *message.mid* )
+    client.ack( *message.mid* , *message.qos* )
 
 
 Constructor Example
--- a/README.rst
+++ b/README.rst
@@ -198,7 +198,7 @@ transport
     "tcp" to use raw TCP.
 
 manual_ack
-    defaults to False, allowing the library to acknowledge messages before
+    defaults to False, allowing the library to acknowledge messages automatically after on_message callback return
     passing them to on_message callback.  When set to True, every message
     must be manually acknowledged by application call to 
     client.ack( *message.mid* , *message.qos* )
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -491,7 +491,8 @@ class Client(object):
     """
 
     def __init__(self, client_id="", clean_session=None, userdata=None,
-                 protocol=MQTTv311, transport="tcp", reconnect_on_failure=True):
+                 protocol=MQTTv311, transport="tcp", reconnect_on_failure=True,
+                 manual_ack=False ):
         """client_id is the unique client id string used when connecting to the
         broker. If client_id is zero length or None, then the behaviour is
         defined by which protocol version is in use. If using MQTT v3.1.1, then
@@ -523,11 +524,20 @@ class Client(object):
 
         Set transport to "websockets" to use WebSockets as the transport
         mechanism. Set to "tcp" to use raw TCP, which is the default.
+
+        Normally, when a message is received, the library automatically
+        acknowledges immediately.  manual_ack=True allows the application to
+        acknowledge receipt after it has completed processing of a message
+        using a the ack() method. This addresses vulnerabilty to message loss
+        if applications fails while processing a message, or while it pending
+        locally.
+
         """
 
         if transport.lower() not in ('websockets', 'tcp'):
             raise ValueError(
                 'transport must be "websockets" or "tcp", not %s' % transport)
+        self._manual_ack = manual_ack
         self._transport = transport.lower()
         self._protocol = protocol
         self._userdata = userdata
@@ -3328,7 +3338,10 @@ class Client(object):
             return MQTT_ERR_SUCCESS
         elif message.qos == 1:
             self._handle_on_message(message)
-            return self._send_puback(message.mid)
+            if self._manual_ack:
+                return MQTT_ERR_SUCCESS
+            else:
+                return self._send_puback(message.mid)
         elif message.qos == 2:
             rc = self._send_pubrec(message.mid)
             message.state = mqtt_ms_wait_for_pubrel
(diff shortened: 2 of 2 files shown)


petersilva:
@ralight not trying to be pushy... but you were very helpful with my last PR (#554) I thought a mild poke might be reasonable ?

petersilva:
@aks  @cclauss  this PR is actually one I care about... nobody seems to be looking at it.  I added the *closes* magic phrase once I learned it from the other PR.

petersilva:
@BBBSnowball @HaraldGustafsson  Does this look like it solves the problem? Perhaps try out the branch and confirm if it works for you?

README.rst
```diff
@@ -197,6 +197,12 @@ transport
     set to "websockets" to send MQTT over WebSockets. Leave at the default of
     "tcp" to use raw TCP.
 
+manual_ack
+    defaults to False, allowing the library to acknowledge messages before
```
PierreF, line 201:
```suggestion
    defaults to False, allowing the library to acknowledge messages automatically after on_message callback return
```

This behavior changed in release 1.6.0 (https://github.com/eclipse/paho.mqtt.python/commit/9a4941efd758ca4231d4f16a8d3951686e24cb92)

src/paho/mqtt/client.py
```diff
@@ -523,11 +524,20 @@ def __init__(self, client_id="", clean_session=None, userdata=None,
 
         Set transport to "websockets" to use WebSockets as the transport
         mechanism. Set to "tcp" to use raw TCP, which is the default.
+
+        Normally, when a message is received, the library automatically
+        acknowledges immediately.  manual_ack=True allows the application to
```
PierreF, line 529:
```suggestion
        acknowledges after on_message callback returns.  manual_ack=True allows the application to
```

akx:
@PierreF Ping – you recently closed #348, mind taking a look at this?

src/paho/mqtt/client.py
```diff
@@ -3328,16 +3338,44 @@ def _handle_publish(self):
...
+
             message.state = mqtt_ms_wait_for_pubrel
             with self._in_message_mutex:
                 self._in_messages[message.mid] = message
+
             return rc
         else:
             return MQTT_ERR_PROTOCOL
 
+    def ack( self, mid, qos ):
```
PierreF, line 3357:
```suggestion
    def ack(self, mid: int, qos: int) -> int:
```

The formatting isn't the expected one (space around parenthesis) and we if could add types it's better.

None are required, because both linter (black for formating and mypy for typing) runs (`tox -e lint`) but only give warning. But if the PR could avoid increasing errors it will be less work in the future :)

src/paho/mqtt/client.py
```diff
@@ -3328,16 +3338,44 @@ def _handle_publish(self):
...
+        """
+        if self._manual_ack :
+            if qos == 1:
+                return self._send_puback(mid)
+            elif qos == 2:
+                return self._send_pubcomp(mid)
+
+        return MQTT_ERR_SUCCESS
+
+    def manual_ack( self, on=False ):
```
PierreF, line 3370:
```suggestion
    def manual_ack_set(self, on):
```

I think naming it `thing_set` is better. It sets an option. `set_manual_ack` sound more natural, but all other setter use `thing_set` (`tls_set`, `reconnect_delay_set`...)

On the non-default value I'm a bit mitigated. For consistency we should have a default. But it feels really odd. You call `manual_ack_set()` and it get disabled ? I think `manual_ack_set` has one flag which should be required, even if that not very consistent with other setter.

PierreF (changes requested):
Could you also add a Signed-off to your commit ? This is a requirement from Eclipse to acknowledge you agree with the ECA.

petersilva:
all good improvments. Thanks @PierreF !

PierreF (approved):

PierreF:
Thank for your contribution.

petersilva:
Thanks!  I think a lot of people are helped by this new function.
I know committters on this repo are overloaded, but it isn't obvious how to help.
If committers want some help, I would be willing, but I don't  see how to get sufficient permission to be helpful.

Issue #493, opened 2020-05-22
SRV connection do not work (with fix)


edwin-oetelaar:
https://github.com/eclipse/paho.mqtt.python/blob/1eec03edf39128e461e6729694cf5d7c1959e5e4/src/paho/mqtt/client.py#L974

Please replace line with parameters instead of wrong positional arguments

  # return self.connect(host, port, keepalive, bind_address, clean_start, properties)
                # Edwin van den Oetelaar fixes this 2020-05-22
                return self.connect(host=host, port=port, keepalive=keepalive, bind_address=bind_address, clean_start=clean_start, properties=properties)

TD22057:
In case this bug isn't clear - this is serious (i.e. breaking) issue.  The connect line in client.py is not passing the arguments through in the correct locations to connect().  As far as I can see, this means args like clean_start don't work at all.

edwin-oetelaar:
I am not sure how to contribute here even more than I already did.
My code running a real project broke, I needed the SRV functions.
I found the problem, debugged it, fixed it and submitted a patch. 
Who is responsible for maintaining this code base? 

Also https://github.com/eclipse/paho.mqtt.python/issues/442#issue-521739758

TD22057:
Any update on this?  This bug breaks the client and is trivial to fix.

and3rson:
Any updates? This really breaks the `connect_srv` and makes it unusable.

edwin-oetelaar:
https://github.com/eclipse/paho.mqtt.python/blob/9782ab81fe7ee3a05e74c7f3e1d03d5611ea4be4/src/paho/mqtt/client.py#L951

this looks still not fixed

@icraggs : please fix what you broke

PR #759, opened 2023-10-23
Fix connect_srv()'s arguments to include bind_port

mijofa:
Fixes this ~3yr old bug https://github.com/eclipse/paho.mqtt.python/issues/493

I think it worth considering reworking this function to use `*args, **kwargs` for these extra arguments, or at least call `connect()` with keyword arguments.
Either one would reduce issues with future failings to manually keep the functions in sync.
But that will likely raise a more intensive workflow discussion, and my goal for this is just to get this 3yr old bug fixed so I can stop using my terrible workarounds.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -913,13 +913,13 @@ class Client(object):
                            bind_address, bind_port, clean_start, properties)
         return self.reconnect()
 
-    def connect_srv(self, domain=None, keepalive=60, bind_address="",
+    def connect_srv(self, domain=None, keepalive=60, bind_address="", bind_port=0,
                     clean_start=MQTT_CLEAN_START_FIRST_ONLY, properties=None):
         """Connect to a remote broker.
 
         domain is the DNS domain to search for SRV records; if None,
         try to determine local domain name.
-        keepalive, bind_address, clean_start and properties are as for connect()
+        keepalive, bind_address, bind_port, clean_start and properties are as for connect()
         """
 
         if HAVE_DNS is False:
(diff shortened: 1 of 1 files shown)


petersilva:
This adds an argument in the middle of an existing entry_point, so anyone using positional parameters will see breakage.  Is there a good readon for that? Can/should we avoid breakage by putting the new parameter at the end?

petersilva:
You need to convince github that you have signed the ECA.

mijofa:
> This adds an argument in the middle of an existing entry_point, so anyone using positional parameters will see breakage. Is there a good readon for that? Can/should we avoid breakage by putting the new parameter at the end?

It is currently completely broken, making this function 100% unusable,
this has been the case for at least 3yrs. (ref: issue https://github.com/eclipse/paho.mqtt.python/issues/493)

I figured since it's already completely broken then might as well stick to consistency with the underlying `connect()` function.
This, however, is what my suggestion of using `*args, **kwargs` for the arguments being passed through directly is meant to kind of resolve and future-proof.
Calling `connect()` with keyword arguments would also avoid future issues with arguments changing,
but that would not allow `connect_srv()` to make use of any arguments added to `connect()` in future.

> You need to convince github that you have signed the ECA.

I don't have an Eclipse account, nor do I want one, I just want this 3yr old bug fixed.
I've pushed a 5min patch and suggested some further improvements,
and now want to wash my hands of this and move on.

I've read the ECA and can certify that:
> The contribution was created in whole or in part by me and I 
> have the right to submit it under the open source license 
> indicated in the file;

and 

> I understand and agree that this project and the contribution 
> are public and that a record of the contribution (including all 
> personal information I submit with it, including my signoff) is 
> maintained indefinitely and may be redistributed consistent with 
> this project or the open source license(s) involved.

PierreF:
Thank for your contribution, but I'm not sure a simple comment that you agree on the ECA is enough for Eclipse process.

While I added typing, I've hit this same issue and fixed it in #791.

I'm closing this PR in favor of #791

PR #757, opened 2023-10-17
address #713: remove claims of python2 support, as it is no longer su…

petersilva:
* Closes #713
* Closes #714

#713 mentions that paho documentation says that it should be compatible with python 2.7.  Someone answered that python2.7 is not supported.

So this PR corrects the documentation, and metadata to remove claims about python 2 support for the current version.

PR #767, opened 2023-12-20
Version compatibility bits (remove old compat code, bump requirement to 3.7+)

akx:
With #757 and 0bc4d41ab51d2364df255e5b7f1033051f7b7bbf, it's clear that this library is not meant to support Pythons older than 3.7, so:

* remove unnecessary Python version check bits
* add the `requires_python` stanza to `setup.py` to mark this as compatible with Python 3.7+.


akx:
@PierreF I see you pushed https://github.com/eclipse/paho.mqtt.python/commit/0bc4d41ab51d2364df255e5b7f1033051f7b7bbf to `master` – does this mean we should target Python 3.7+?

PierreF:
BTW: could you add a Signed-Off to your commit ? This is a requirement from Eclipse to acknowledge you agree with the ECA.

akx:
@PierreF Added signoffs and bumped the references to 3.6 to 3.7.

src/paho/mqtt/client.py
```diff
@@ -81,21 +81,10 @@
...
-except NameError:
-    BlockingIOError  = IOError
-
 MQTTv31 = 3
 MQTTv311 = 4
 MQTTv5 = 5
 
-if sys.version_info[0] >= 3:
-    # define some alias for python2 compatibility
-    unicode = str
```
PierreF, line 96: This cause some issue, we had reference to  `unicode` that need to be changed to `str`

The same happen for `basestring` that should be changed to `str` also.

akx: Done. Some tests still don't pass, but I think that's the case on `master` to.

PR #771, opened 2023-12-21
CI: Test on newer Pythons; enable Python dev mode

akx:
See https://docs.python.org/3.12/library/devmode.html#devmode for what `PYTHONDEVMODE` does.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -6,13 +6,17 @@ jobs:
       fail-fast: false
       max-parallel: 4
       matrix:
-        python: [3.7, 3.8, 3.9]
+        python: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python }}
+          cache: pip
+          cache-dependency-path: |
+            tox.ini
+            setup.py
       - run: pip install tox
       - if: matrix.python == '3.7'
         run: TOXENV=py37 tox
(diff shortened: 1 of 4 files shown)


akx:
@PierreF Rebased, ready for review.

.github/workflows/tox.yml
```diff
@@ -6,17 +6,27 @@ jobs:
...
       - if: matrix.python == '3.8'
         run: TOXENV=py38 tox
       - if: matrix.python == '3.9'
         run: TOXENV=py39 tox
+      - if: matrix.python == '3.10'
+        run: TOXENV=py310 tox
+      - if: matrix.python == '3.11'
+        run: TOXENV=py311 tox
+      - if: matrix.python == '3.12'
+        run: TOXENV=py312 tox
```
cclauss, line 32: Let's use the `- run: tox -e py` as discussed in the GHA docs and #777

tox.ini
```diff
@@ -11,8 +11,10 @@ allowlist_externals =
     make
 commands =
     ruff .
-    python setup.py test
+    pytest
     make -C test test
+env =
+    PYTHONDEVMODE=1
```
cclauss, line 17: How does this work?

cclauss (approved):
Love the addition of `PYTHONDEVMODE` -- Cool!

PierreF:
The PR looks fine, we just broke "python setup.py test" command.
I'm not sure whether is really used (I think the the release process, which might use use). Anyway, I'll have a look at #774 which should both cause few conflict with this PR and remove "setup.py" completely.

cclauss:
https://setuptools.pypa.io/en/latest/deprecated/commands.html
> running `python setup.py` directly as a script is considered __deprecated__. This also means that users should avoid running commands directly via `python setup.py <command>`.

akx:
Yeah, breaking `setup.py test` was on purpose – there's no reference to it anywhere in the repo, so I figured it wasn't really needed (anymore)...

PierreF:
Confirmed that setup.py is not used for release, at least not how I do release (which I should document somewhere).

My process is currently:
* testing by calling tox (inside a Docker image python:3.x)
* testing linter by calling tox
* tag & push tag
* build and upload to PyPi (using make upload)

Therefor this PR seems fine for me.

akx:
> My process is currently: ...

If it helps, the procedure I've built for python-babel is pretty slick these days – everything is done on GitHub Actions, and I just okay the release (via the `release` GHA environment requiring an accept gate).

PierreF:
That true that making a release on CI is better than locally, as it ensure your local environment don't affect the release. I take not of your workflow and will consider it.

PR #718, opened 2023-04-16
GitHub Action to lint Python code with ruff

cclauss:
Failing tests are fixed in #712

[Ruff](https://beta.ruff.rs/) supports [over 500 lint rules](https://beta.ruff.rs/docs/rules) and can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [yesqa](https://github.com/asottile/yesqa), [eradicate](https://pypi.org/project/eradicate/), [pyupgrade](https://pypi.org/project/pyupgrade/), and [autoflake](https://pypi.org/project/autoflake/), all while executing (in Rust) tens or hundreds of times faster than any individual tool.

The ruff Action uses minimal steps to run in ~5 seconds, rapidly providing intuitive GitHub Annotations to contributors.

image

PR #776, opened 2023-12-23
Add pre-commit plus a GItHub Action to run it

cclauss:
As discussed at https://github.com/eclipse/paho.mqtt.python/pull/718#discussion_r1432593177

@akx Your review, please.

Using https://pre-commit.ci is even better than using GitHub Actions for pre-commit.

% `pre-commit run --all-files`
```
[INFO] Initializing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Initializing environment for https://github.com/charliermarsh/ruff-pre-commit.
[INFO] Initializing environment for https://github.com/abravalheri/validate-pyproject.
[INFO] Initializing environment for https://github.com/abravalheri/validate-pyproject:.[all].
[INFO] Installing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/charliermarsh/ruff-pre-commit.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/abravalheri/validate-pyproject.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
check builtin type constructor use.......................................Passed
check toml...............................................................Passed
check xml............................................(no files to check)Skipped
check yaml...............................................................Passed
ruff.....................................................................Passed
Validate pyproject.toml..................................................Passed
```

### DIFF
--- a/.github/workflows/precommit.yml
+++ b/.github/workflows/precommit.yml
@@ -15,6 +15,6 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: 3.x
-      - run: pip install pre-commit
+      - use: pre-commit/action@v3.0.0
       - run: pre-commit --version
       - run: pre-commit run --all-files
--- a/.github/workflows/precommit.yml
+++ b/.github/workflows/precommit.yml
@@ -17,5 +17,4 @@ jobs:
           python-version: 3.x
       - run: pip install pre-commit
       - run: pre-commit --version
-      - run: pre-commit install
       - run: pre-commit run --all-files
--- a/.github/workflows/precommit.yml
+++ b/.github/workflows/precommit.yml
@@ -15,6 +15,4 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: 3.x
-      - use: pre-commit/action@v3.0.0
-      - run: pre-commit --version
-      - run: pre-commit run --all-files
+      - uses: pre-commit/action@v3.0.0
--- a/.github/workflows/precommit.yml
+++ b/.github/workflows/precommit.yml
@@ -1,0 +1,21 @@
+# https://pre-commit.com
+# This GitHub Action assumes that the repo contains a valid .pre-commit-config.yaml file.
+# Using pre-commit.ci is even better that using GitHub Actions for pre-commit.
+name: pre-commit
+on:
+  pull_request:
+    branches: [master]
+  push:
+  workflow_dispatch:
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: 3.x
+      - run: pip install pre-commit
+      - run: pre-commit --version
+      - run: pre-commit install
+      - run: pre-commit run --all-files
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,39 @@
+# Learn more about this config here: https://pre-commit.com/
+
+# To enable these pre-commit hooks run:
+# `brew install pre-commit` or `python3 -m pip install pre-commit`
+# Then in the project root directory run `pre-commit install`
+
+# default_language_version:
+#   python: python3
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.5.0
+    hooks:
+      - id: check-builtin-literals
+      # - id: check-executables-have-shebangs
+      # - id: check-shebang-scripts-are-executable
+      - id: check-toml
+      - id: check-xml
+      - id: check-yaml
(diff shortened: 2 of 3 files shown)


.github/workflows/precommit.yml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: 3.x
+      - run: pip install pre-commit
+      - run: pre-commit --version
+      - run: pre-commit install
+      - run: pre-commit run --all-files
```
akx, line 21:
```suggestion
      - uses: actions/setup-python@v5
        with:
          python-version: 3
      - uses: pre-commit/action@v3.0.0
```

.github/workflows/precommit.yml
```diff
@@ -0,0 +1,21 @@
...
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: 3.x
+      - run: pip install pre-commit
+      - run: pre-commit --version
+      - run: pre-commit install
```
akx, line 20: Installing the Git hook is unnecessary in CI :)

akx (commented):
This really boils down to maintainer (@PierreF) preference. #770 already set up running Ruff with `tox`, which is also fine. Left a comment about https://github.com/pre-commit/action in case you weren't aware of it.

.github/workflows/precommit.yml
cclauss: https://github.com/pre-commit/action is in maintenance-only mode and will not be accepting new features so I tend not to use it.

.github/workflows/precommit.yml
cclauss:
```suggestion
```

.github/workflows/precommit.yml
akx: I know it's in maintenance-only mode, but it works – it's also just, well, [20 lines](
https://github.com/pre-commit/action/blob/main/action.yml).

I don't prefer pre-commit.ci because it's an one-man project (and for projects where "auto-fix PRs" is enabled, they tend to mess with my workflow).

cclauss: If auto-fix PRs are messing up your workflow then you do not have pre-commit properly installed locally.

.github/workflows/precommit.yml
```diff
@@ -0,0 +1,20 @@
...
+  workflow_dispatch:
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: 3.x
+      - run: pip install pre-commit
```
PierreF, line 18: To my understanding we could use
```suggestion
      - use: pre-commit/action@v3.0.0
```

Which mostly do the same (pip install and pre-commit run) but also add a cache.

Even if only in maintenance, if the version 3.0.0 had an issue, we could just copy their actions.yaml and fix it here.

PierreF (approved):
This add another tools for checking code, but I think it's okay:
* tox -e py (or just tox / tox -e py38): is here to run Python tests. And might took some time
* tox -e lint will be kept for slow linter
* pre-commit will be used for faster linter

Those linters/tests could still run locally and in Github actions. Which fit the needs of everyone:
* the one that want to run it locally (like me) before pushing to Github
* all PRs and master is also automatically checked by Github

.github/workflows/precommit.yml
```diff
@@ -0,0 +1,20 @@
...
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: 3.x
+      - use: pre-commit/action@v3.0.0
+      - run: pre-commit --version
+      - run: pre-commit run --all-files
```
PierreF, line 20:
```suggestion
      - uses: pre-commit/action@v3.0.0
```

Oups, I made a typo in my previous suggestion

PierreF:
The error isn't added by the PR but already present on master. I will fix it once merged.

akx (commented):
Weak review comment, can be addressed later as well: If this is adopted (and I'm for it, in general), `tox -e lint` should just run `pre-commit run --all-files` to avoid drift between multiple lint styles.

.github/workflows/precommit.yml
akx: @cclauss Well, for a repo like this where every commit should have a `--signoff` to adhere to a CLA, the pre-commit commit wouldn't be appropriate, etc. 😁

PR #779, opened 2023-12-23
Enable more Ruff rules (incl. those `ignore`d previously)

akx:


akx:
🔔 @cclauss 😁

PierreF:
I think there will be a conflict between this PR and #776 . #776 is currently mergeable (no conflict), so it should probably be better to merge #776 first. @akx, you were asked for a review on that PR that why I wait to merge it.

akx:
Draft for the time being – #776 reorganized pyproject.toml in a way that caused rather annoying conflicts so I'll re-check things once more.

akx:
@PierreF Rebased, ready for review.

PierreF:
Thank ! That's a good cleanup

PR #780, opened 2023-12-23
Convert homespun tests to conventional tests

### LINKED ISSUE #768: Convert `test/Makefile` tests to regular py.test tests
> Those tests should be converted to be regular `py.test` tests IMO, they now seem to rely on some _very_ homespun code (e.g. some Python code in `.test` files that maybe somehow relates to the `.py` files...).
>
> They're also not covered by any code coverage so long as that's happening.

_Originally posted by @akx in https://github.com/eclipse/paho.mqtt.python/pull/718#discussion_r1432723019_

akx:
This PR converts the homespun `test/` code to (what I think is) pretty idiomatic Python/Pytest code.

I think I got everything right, the original code was pretty hairy. 😂 

Would probably be even better if the client code was included in the tests themselves to avoid having to hop between files.

Fixes #768


tests/lib/clients/01-asyncio.py
```diff
@@ -91,6 +87,7 @@ async def main(self):
         await self.disconnected
         rc = 0
 
+
 rc = 1
 loop = asyncio.get_event_loop()
 loop.run_until_complete(AsyncMqttExample(loop).main())
```
PierreF, line 93: (i can't comment on right line). I think `exit(rc)` should be replaced by `assert rc == 0`. Without that pytest will succeed even when test fail.

PierreF: Well actually it's not enough. I'm trying by changing  line 88 to `rc = 1` (which make the test failing). pytest still works without error.

PierreF: By move those line inside a function named test is make test being run:
```
def test():
    loop = asyncio.get_event_loop()
    loop.run_until_complete(AsyncMqttExample(loop).main())
    loop.close()
    assert rc == 0
```

But we are missing to run the fake server which cause a timeout error

PierreF: After a second read, I think we just need to add `assert proc.wait() == 0` in test_01_asyncio.py

tests/lib/test_01_asyncio.py
```diff
@@ -45,12 +40,5 @@
     conn.send(puback_packet_in)
 
     paho_test.expect_packet(conn, "disconnect", disconnect_packet)
-    rc = 0
 
     conn.close()
```
PierreF, line 44:
```suggestion
    conn.close()
    assert proc.wait() == 0
```

tests/lib/test_01_asyncio.py
```diff
@@ -23,12 +19,11 @@
 
 disconnect_packet = paho_test.gen_disconnect()
 
-sock = paho_test.create_server_socket()
 
-client = context.start_client()
+def test_01_asyncio(server_socket, start_client):
+    start_client("01-asyncio.py")
```
PierreF, line 24:
```suggestion
    proc = start_client("01-asyncio.py")
```

cclauss:
pytest, not py.test since version 3.0

https://www.youtube.com/watch?v=JIAXHEqjuNA

akx:
> pytest, not py.test since version 3.0

old naming dies hard 😂

akx:
Ah, I'll improve this a little more – we'll want to be sure the client process didn't die with an exception and so on...

akx:
Phew, this PR underwent a bit of a metamorphosis since last night, but I think it's much, much better now. We will now be able to automagically catch exceptions from the clients, and there's (almost) no playing with return codes and so on.

A bunch of actually unused/no-op code also got removed – a 723 line removal is pretty good in my books 😁

tests/paho_test.py
```diff
@@ -25,13 +24,26 @@ class TestError(Exception):
     def __init__(self, message="Mismatched packets"):
         self.message = message
 
+
+def bind_to_any_free_port(sock) -> int:
```
PierreF, line 28: ❤️ I will no longer have strange error when I run a real broker on port 1888 and run paho test.

tests/lib/conftest.py
```diff
@@ -0,0 +1,71 @@
...
+        client_path = clients_path / name
+        if not client_path.exists():
+            raise FileNotFoundError(client_path)
+        env = dict(
+            os.environ,
+            PAHO_SSL_PATH=str(ssl_path),
+            PYTHONPATH=f"{tests_path}{os.pathsep}{os.environ.get('PYTHONPATH', '')}",
+        )
+        assert 'PAHO_SERVER_PORT' in env, "PAHO_SERVER_PORT must be set in the environment when starting a client"
+        # TODO: it would be nice to run this under `coverage` too!
```
PierreF, line 57: We might be able to have coverage done it instead of a subprocess we start a thread. But this is probably easy to said and harder to implement. Anyway I think it not in the scope of this PR.

PierreF (approved):
Great works ! That huge works, thank for your time.

tests/lib/conftest.py
akx: `pytest-cov` actually hooks things up so subprocesses work OOTB, which was nice to notice!

akx:
> Great works ! That huge works, thank for your time.

No problem, it was a nice brain workout while watching Christmas movies and drinking mulled wine and cognac... 😂

PR #781, opened 2023-12-24
(Kind of) fix mqttv5 tests

### LINKED ISSUE #769: CI: also run mqtt5 tests
Currently the `test_mqtt5` tests aren't being run in CI, because the requisite submodule isn't getting cloned into place.

(The tests don't _work_, either...)

akx:
Fixes #769.

The https://github.com/eclipse/paho.mqtt.testing repo is not in a pretty state, but fixing it is maybe a bit out-of-scope for this PR.

I think some of the tests may be flaky and timeout-dependent, but that's another issue to fix...

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -10,6 +10,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


PierreF:
Thank once more for your work ❤️

PR #783, opened 2023-12-25
Use f-strings where sensible

akx:
This PR replaces various string formatting operations with Python 3.6+ f-strings, for performance and brevity.

Guided by [Flynt](https://github.com/ikamensh/flynt/) and Ruff's relevant lints, checked by hand.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,6 +79,7 @@ extend-select = [
     "F63",
     "F7",
     "F82",
+    "FLY",  # flynt
     "I",
     "S", # Bandit
     "UP",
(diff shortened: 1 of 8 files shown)


cclauss:
Do ruff [`EM` rules](https://docs.astral.sh/ruff/rules/#flake8-errmsg-em) object to some of these uses?

akx:
@cclauss Not any more than they would with other formatting. I don't really agree with the `EM` rules because they make the code longer and more cumbersome to read. Henry's point [is](https://pypi.org/project/flake8-errmsg/)

> If this is longer or more complex, the duplication can be quite confusing for a user unaccustomed to reading tracebacks.

and I would assume people who use MQTT should be accustomed to reading tracebacks.

PierreF (approved):

cclauss:
@akx git conflicts here.

akx:
@cclauss Thanks, rebased.

PR #784, opened 2023-12-27
Remove dead testing code

akx:
This PR removes a bunch of dead testing code (that was added in a155a876c3a8422d9f88d555dcfc5636817c8e76 but never used thereafter?) and splits the `tests.paho_test` module into slightly more manageable pieces.

It would be nice to refactor things altogether so there wouldn't be duplicate packet generation code in the library itself and separately again in the testing code, but by the looks of it, that's going to take some serious untangling.


PierreF (approved):

akx:
@PierreF Rebased 👍

PR #786, opened 2024-01-01
Add more linter

PierreF:
Add black, codespell and mypy.

mypy is only partially added, mostly to client.py.

PR #791, opened 2024-01-03
Add typing with mypy

PierreF:
Split of #786 : contains add of mypy on client.py


src/paho/mqtt/client.py
```diff
@@ -19,6 +19,7 @@
 
 import base64
```
akx, line 20: It could be a good idea to consider `from __future__ import annotations`, so you wouldn't need the stringly-typed future references (e.g. `Callable["Client"]` could be `Callable[Client]`).

That will also allow you to use modern type annotation syntax even if the target Python version might not necessarily support it (e.g. `Foo | None` instead of `Optional[Foo]`), as type checkers do understand the syntax. Ruff has a rule or two to do that auto-conversion.

src/paho/mqtt/client.py
```diff
@@ -131,25 +161,49 @@
...
-MQTT_ERR_TLS = 8
-MQTT_ERR_PAYLOAD_SIZE = 9
-MQTT_ERR_NOT_SUPPORTED = 10
-MQTT_ERR_AUTH = 11
-MQTT_ERR_ACL_DENIED = 12
-MQTT_ERR_UNKNOWN = 13
-MQTT_ERR_ERRNO = 14
-MQTT_ERR_QUEUE_SIZE = 15
-MQTT_ERR_KEEPALIVE = 16
+class MQTTErrorCode(enum.IntEnum):
```
akx, line 166: Maybe this should be in a new module, say, `mqtt.enums`? The integer aliases could remain here for compatibility.

PierreF:
Tests are a bit flappy (not related to this PR, it's also happen on master or other PR). @akx did you also had unstable tests result for PR you created recently (i.e. did you re-run failed tests until they succeed) ? Because if you didn't had to re-run failed test, tests seems to fail more often recently.

src/paho/mqtt/client.py
```diff
@@ -30,6 +31,7 @@
 import struct
 import threading
 import time
+import typing
```
akx, line 34: I would probably `from typing import ...` all of these to make things easier to read?

src/paho/mqtt/client.py
```diff
@@ -222,7 +321,9 @@ def connack_string(connack_code):
         return "Connection Refused: unknown reason."
 
 
-def base62(num, base=string.digits + string.ascii_letters, padding=1):
+def base62(
+    num: int, base: str = string.digits + string.ascii_letters, padding: int = 1
```
akx, line 325: I'd add the trailing comma already, so this would reformat to
```suggestion
    num: int,
    base: str = string.digits + string.ascii_letters,
    padding: int = 1,
```

src/paho/mqtt/client.py
```diff
@@ -66,10 +68,38 @@
...
+        qos: int
+        pos: int
+        to_process: int
+        packet: bytes
+        info: typing.Optional["MQTTMessageInfo"]
+
+
 MQTTv31 = 3
 MQTTv311 = 4
 MQTTv5 = 5
```
akx, line 105: These should probably also be an enum?

src/paho/mqtt/client.py
```diff
@@ -675,12 +800,19 @@ def _reset_sockets(self, sockpair_only=False):
...
+        userdata: typing.Any = None,
+    ) -> None:
         self._reset_sockets()
 
-        self.__init__(client_id, clean_session, userdata)
+        self.__init__(client_id, clean_session, userdata)  # type: ignore[misc]
 
-    def ws_set_options(self, path="/mqtt", headers=None):
+    def ws_set_options(
+        self, path: str = "/mqtt", headers: typing.Optional[WebSocketHeaders] = None
```
akx, line 814:
```suggestion
        self,
        path: str = "/mqtt",
        headers: typing.Optional[WebSocketHeaders] = None,
```

src/paho/mqtt/client.py
```diff
@@ -700,7 +832,9 @@ def ws_set_options(self, path="/mqtt", headers=None):
                 raise ValueError(
                     "'headers' option to ws_set_options has to be either a dictionary or callable")
 
-    def tls_set_context(self, context=None):
+    def tls_set_context(
+        self, context: typing.Optional["ssl.SSLContext"] = None
```
akx, line 836:
```suggestion
        self,
        context: typing.Optional["ssl.SSLContext"] = None,
```

src/paho/mqtt/client.py
```diff
@@ -870,11 +1013,21 @@ def enable_logger(self, logger=None):
...
+    def connect(
+        self,
+        host: str,
+        port: int = 1883,
+        keepalive: int = 60,
+        bind_address: str = "",
+        bind_port: int = 0,
+        clean_start: typing.Union[
+            bool, "Literal[3]"
+        ] = MQTT_CLEAN_START_FIRST_ONLY,  # type: ignore
```
akx, line 1028: Should `typing.Union[bool, Literal[3]]` maybe be refactored into a type alias?

src/paho/mqtt/client.py
```diff
@@ -1042,16 +1214,19 @@ def reconnect(self):
                 if not self.suppress_exceptions:
                     raise
 
-        sock = self._create_socket_connection()
```
akx, line 1045: The changes from here up to where `self._sock` gets assigned smell like this should be refactored to a separate `self._create_socket()` function (which would probably type more nicely).

src/paho/mqtt/client.py
```diff
@@ -1315,11 +1508,12 @@ def username_pw_set(self, username, password=None):
 
         # [MQTT-3.1.3-11] User name must be UTF-8 encoded string
         self._username = None if username is None else username.encode('utf-8')
-        self._password = password
-        if isinstance(self._password, str):
-            self._password = self._password.encode('utf-8')
+        if isinstance(password, str):
+            self._password = password.encode('utf-8')
+        else:
+            self._password = password
```
akx, line 1514: This idiom should probably be refactored into a free helper function. (I've seen it called `force_bytes()` in e.g. Django, with `force_str()` being the counterpart.)

src/paho/mqtt/client.py
```diff
@@ -2176,19 +2420,28 @@ def on_socket_open(self, func):
...
             on_socket_open = self.on_socket_open
 
         if on_socket_open:
             with self._in_callback_mutex:
+                if self._sock is None:
+                    self._easy_log(
+                        MQTT_LOG_ERR,
+                        "socket() is None in _call_socket_open",
+                    )
+                    return
```
akx, line 2443: This sounds like a "should-never-happen" sort of bug – maybe `_call_socket_open()`should be called by the client internally with the newly opened socket as an argument, so it wouldn't need `self._sock`?

src/paho/mqtt/client.py
```diff
@@ -2407,20 +2680,20 @@ def _packet_read(self):
...
-                return MQTT_ERR_CONN_LOST
+                return MQTTErrorCode.MQTT_ERR_CONN_LOST
             else:
                 if len(command) == 0:
-                    return MQTT_ERR_CONN_LOST
-                command, = struct.unpack("!B", command)
-                self._in_packet['command'] = command
+                    return MQTTErrorCode.MQTT_ERR_CONN_LOST
+                command_value, = struct.unpack("!B", command)
+                self._in_packet['command'] = command_value
```
akx, line 2696: Or just
```suggestion
                self._in_packet['command'] = command[0]
```
since `command` is a `bytes`, and indexing into `bytes` gives you integers (and `B` doesn't need endianness):

```
>>> buf = b"@"
>>> struct.unpack("!B", buf)[0]
64
>>> buf[0]
64
```

src/paho/mqtt/client.py
```diff
@@ -2430,26 +2703,26 @@ def _packet_read(self):
...
                         MQTT_LOG_ERR, 'failed to receive on socket: %s', err)
-                    return MQTT_ERR_CONN_LOST
+                    return MQTTErrorCode.MQTT_ERR_CONN_LOST
                 else:
                     if len(byte) == 0:
-                        return MQTT_ERR_CONN_LOST
-                    byte, = struct.unpack("!B", byte)
-                    self._in_packet['remaining_count'].append(byte)
+                        return MQTTErrorCode.MQTT_ERR_CONN_LOST
+                    byte_value, = struct.unpack("!B", byte)
```
akx, line 2714: As above:
```suggestion
                    byte_value = byte[0]
```

src/paho/mqtt/client.py
```diff
@@ -2537,15 +2811,19 @@ def _packet_write(self):
                                     if not self.suppress_exceptions:
                                         raise
 
-                        packet['info']._set_as_published()
+                        # TODO: Something is odd here. I don't see why packet["info"] can't be None.
+                        # A packet could be produced by _handle_connack with qos=0 and no info
+                        # (around line 3645). Ignore the mypy check for now but I feel their is a bug
```
akx, line 2816:
```suggestion
                        # (around line 3645). Ignore the mypy check for now but I feel there is a bug
```

src/paho/mqtt/client.py
```diff
@@ -2595,58 +2873,60 @@ def _check_keepalive(self):
...
+    def _mid_generate(self) -> int:
         with self._mid_generate_mutex:
             self._last_mid += 1
             if self._last_mid == 65536:
                 self._last_mid = 1
             return self._last_mid
 
     @staticmethod
-    def _topic_wildcard_len_check(topic):
+    def _topic_wildcard_len_check(topic: bytes) -> MQTTErrorCode:
```
akx, line 2890: It would probably be more pythonic to just raise from here instead of return an error code (that's only ever used internally, one hopes) – or at least return a `bool`.

src/paho/mqtt/client.py
```diff
@@ -3241,24 +3555,24 @@ def _handle_connack(self):
...
             return rc
         elif result > 0 and result < 6:
-            return MQTT_ERR_CONN_REFUSED
+            return MQTTErrorCode.MQTT_ERR_CONN_REFUSED
         else:
-            return MQTT_ERR_PROTOCOL
+            return MQTTErrorCode.MQTT_ERR_PROTOCOL
 
-    def _handle_disconnect(self):
+    def _handle_disconnect(self) -> "Literal[MQTTErrorCode.MQTT_ERR_SUCCESS]":
```
akx, line 3575: If the only value this function can return is this, should it just return None instead?

src/paho/mqtt/client.py
```diff
@@ -3275,9 +3589,9 @@ def _handle_disconnect(self):
 
         self._loop_rc_handle(reasonCode, properties)
 
-        return MQTT_ERR_SUCCESS
+        return MQTTErrorCode.MQTT_ERR_SUCCESS
 
-    def _handle_suback(self):
+    def _handle_suback(self) -> "Literal[MQTTErrorCode.MQTT_ERR_SUCCESS]":
```
akx, line 3594: Ditto – sounds like this shouldn't return a value.

src/paho/mqtt/client.py
```diff
@@ -3494,25 +3806,29 @@ def _handle_pubrec(self):
...
-                reasoncodes.append(ReasonCodes(UNSUBACK >> 4, identifier=c))
-            if len(reasoncodes) == 1:
-                reasoncodes = reasoncodes[0]
+                reasoncodes_list.append(ReasonCodes(UNSUBACK >> 4, identifier=c))
+
+            reasoncodes: typing.Union[
+                ReasonCodes, typing.List[ReasonCodes]
+            ] = reasoncodes_list
+            if len(reasoncodes_list) == 1:
+                reasoncodes = reasoncodes_list[0]
```
akx, line 3831: Maybe add a TODO comment to remind someone to clean this up? It doesn't look very nice now. (I think `ReasonCodes` and `PacketTypes` should be named in the singular, for one...)

src/paho/mqtt/publish.py
```diff
@@ -63,9 +93,19 @@ def _on_publish(client, userdata, mid):
...
-             transport="tcp", proxy_args=None):
+def multiple(
+    msgs: "MessagesList",
+    hostname: str = "localhost",
+    port: int = 1883,
+    client_id: str = "",
+    keepalive: int = 60,
+    will: typing.Optional["MessageDict"] = None,
+    auth: typing.Optional["AuthParamater"] = None,
+    tls: typing.Optional["TLSParamater"] = None,
```
akx, line 104: Typos? `Paramater`?

src/paho/mqtt/publish.py
```diff
@@ -172,9 +213,22 @@ def multiple(msgs, hostname="localhost", port=1883, client_id="", keepalive=60,
...
+    payload: paho.PayloadType = None,
+    qos: int = 0,
+    retain: bool = False,
+    hostname: str = "localhost",
+    port: int = 1883,
+    client_id: str = "",
+    keepalive: int = 60,
+    will: typing.Optional["MessageDict"] = None,
+    auth: typing.Optional["AuthParamater"] = None,
+    tls: typing.Optional["TLSParamater"] = None,
```
akx, line 227: Ditto, typos?

tests/lib/clients/03-publish-b2c-qos1.py
```diff
@@ -10,7 +10,7 @@ def on_message(mqttc, obj, msg):
     assert msg.topic == "pub/qos1/receive", f"Invalid topic: ({msg.topic})"
     assert msg.payload == expected_payload, f"Invalid payload: ({msg.payload})"
     assert msg.qos == 1, f"Invalid qos: ({msg.qos})"
-    assert msg.retain is not False, f"Invalid retain: ({msg.retain})"
+    assert not msg.retain, f"Invalid retain: ({msg.retain})"
```
akx, line 13: Isn't this the inverse test c.f. before?

tests/test_client.py
```diff
@@ -316,3 +316,44 @@ def callback2(client, userdata, msg):
         assert userdata['on_message'] == 1
         assert userdata['callback1'] == 1
         assert userdata['callback2'] == 2
+
+
+class Test_compatibility:
+    """
+    Few test for backward compatibility
```
akx, line 323:
```suggestion
    Some tests for backward compatibility
```

tox.ini
```diff
@@ -16,7 +15,8 @@ env =
 
 [testenv:lint]
 deps =
-    -e .
+    -e .[proxy]
+    dnspython
```
akx, line 19: Why is this required? 🤔

akx (changes requested):
Some softer suggestions and some actual bug-looking things here.

src/paho/mqtt/client.py
PierreF: I didn't known this one. It's really cool.

PierreF:
Thank for the review. I've only fixed few of your feedback but it's bed time now, I'll continue tomorrow.

tests/lib/clients/03-publish-b2c-qos1.py
PierreF: Yes, but I think the test was wrong.
* We don't produce a message with retain flag in this test, so it make sense that msg.retain is False
* Previously, msg.retain was an interger so `msg.retain is not False` was always true.
* I've checked and the value was `0`. So it seems expected to assert the value `False`

tox.ini
PierreF: I've added them so mypy could check types. This isn't useful (currently) for proxy since it's done have typing.
Bit It help with dnspython for checking type with call to `dns.resolver`

src/paho/mqtt/client.py
PierreF: The issue is that changing this will be a breaking change since it's used in on_subscribe callback.

src/paho/mqtt/client.py
PierreF: Yes, I didn't changed them because it would break the signature (breaking change)... but it's a private function, we are not going to have backward compatibility with private function :)

src/paho/mqtt/enums.py
```diff
@@ -0,0 +1,88 @@
+import enum
+
+
+# Error values
```
akx, line 4: This comment doesn't seem to add much value :)
```suggestion
```

src/paho/mqtt/enums.py
```diff
@@ -0,0 +1,88 @@
...
+    PUBCOMP = 0x70
+    SUBSCRIBE = 0x80
+    SUBACK = 0x90
+    UNSUBSCRIBE = 0xA0
+    UNSUBACK = 0xB0
+    PINGREQ = 0xC0
+    PINGRESP = 0xD0
+    DISCONNECT = 0xE0
+    AUTH = 0xF0
+
```
akx, line 48:
```suggestion


```

src/paho/mqtt/enums.py
```diff
@@ -0,0 +1,88 @@
...
+class PahoLogLevel(enum.IntEnum):
```
akx, line 49: Why is this `PahoLogLevel` when none of the others are `Paho...`? There isn't a `LogLevel` enum in `logging`, so I guess this could as well be
```suggestion
class LogLevel(enum.IntEnum):
```

src/paho/mqtt/enums.py
```diff
@@ -0,0 +1,88 @@
...
+    CONNACK_REFUSED_SERVER_UNAVAILABLE = 3
+    CONNACK_REFUSED_BAD_USERNAME_PASSWORD = 4
+    CONNACK_REFUSED_NOT_AUTHORIZED = 5
+
+
+class ConnectionState(enum.IntEnum):
+    mqtt_cs_new = 0
+    mqtt_cs_connected = 1
+    mqtt_cs_disconnecting = 2
+    mqtt_cs_connect_async = 3
```
akx, line 70: These should maybe also be `UPPER_SNAKE_CASE`? The compatibility aliases in `client.py` can of course remain lower-case.

src/paho/mqtt/publish.py
```diff
@@ -18,15 +18,49 @@
 situation where you have a single/multiple messages you want to publish to a
 broker, then disconnect and nothing else is required.
 """
+from __future__ import annotations
 
 import collections
+import typing
 from collections.abc import Iterable
+from typing import Any, List, Tuple, Union
```
akx, line 26: Might be good to import the `typing` one way only; I'd suggest the `from typing` way.

tests/lib/clients/03-publish-b2c-qos1.py
akx: Maybe this change should be done in a separate PR, then? I'm afraid it will confuse future code archeologists terribly if this happens in a commit that says "Add typing".

tests/test_client.py
```diff
@@ -316,3 +316,44 @@ def callback2(client, userdata, msg):
         assert userdata['on_message'] == 1
         assert userdata['callback1'] == 1
         assert userdata['callback2'] == 2
+
+
+class Test_compatibility:
```
akx, line 321:
```suggestion
class TestCompatibility:
```

src/paho/mqtt/client.py
akx: Probably shouldn't import `typing` in two ways here either.

src/paho/mqtt/client.py
```diff
@@ -3740,7 +3968,18 @@ def _get_proxy(self):
...
 
-    def _create_socket_connection(self):
+    def _create_socket(self) -> SocketLike:
+        tcp_sock = self._create_socket_connection()
+        sock = self._create_ssl_socket_if_enable(tcp_sock)
+
+        if self._transport == "websockets":
+            sock.settimeout(self._keepalive)
+            return WebsocketWrapper(sock, self._host, self._port, self._ssl,
+                                    self._websocket_path, self._websocket_extra_headers)
```
akx, line 3978: Would maybe read easier as 
```suggestion
            return WebsocketWrapper(
                sock,
                self._host,
                self._port,
                self._ssl,
                self._websocket_path,
                self._websocket_extra_headers,
           )
```
(and I'd also consider adding kwargs, so it's harder to accidentally pass things in the wrong order)

src/paho/mqtt/client.py
```diff
@@ -3740,7 +3968,18 @@ def _get_proxy(self):
         # None to indicate that the connection should be handled normally
         return None
 
-    def _create_socket_connection(self):
+    def _create_socket(self) -> SocketLike:
+        tcp_sock = self._create_socket_connection()
+        sock = self._create_ssl_socket_if_enable(tcp_sock)
```
akx, line 3973: I'm not sure the `_if_enable` is a good idea - I'd just do

```suggestion
        if self._ssl:
            sock = self._ssl_wrap_socket(tcp_sock)
```
or similar? That would also simplify the return type of that function.

src/paho/mqtt/client.py
```diff
@@ -3750,6 +3989,48 @@ def _create_socket_connection(self):
...
+        except ValueError:
+            # Python version requires SNI in order to handle server_hostname, but SNI is not available
+            ssl_sock = self._ssl_context.wrap_socket(
+                tcp_sock,
+                do_handshake_on_connect=False,
+            )
+        else:
+            # If SSL context has already checked hostname, then don't need to do it again
+            if (hasattr(self._ssl_context, 'check_hostname') and
+                    self._ssl_context.check_hostname):  # type: ignore
```
akx, line 4021:
```suggestion
            if getattr(self._ssl_context, 'check_hostname', False):  # type: ignore
```

src/paho/mqtt/client.py
```diff
@@ -520,8 +636,9 @@ def __init__(self, client_id="", clean_session=None, userdata=None,
         self._transport = transport.lower()
         self._protocol = protocol
         self._userdata = userdata
-        self._sock = None
-        self._sockpairR, self._sockpairW = (None, None,)
+        self._sock: socket.socket | WebsocketWrapper | ssl.SSLSocket | None = None
```
akx, line 639:
```suggestion
        self._sock: SocketLike | None = None
```

src/paho/mqtt/client.py
```diff
@@ -3084,46 +3308,48 @@ def _packet_handle(self):
...
+            return MQTTErrorCode.MQTT_ERR_PROTOCOL
 
         if self._protocol == MQTTv5:
             (flags, result) = struct.unpack(
                 "!BB", self._in_packet['packet'][:2])
             if result == 1:
                 # This is probably a failure from a broker that doesn't support
                 # MQTT v5.
-                reason = 132 # Unsupported protocol version
+                reason = ReasonCodes(CONNACK >> 4, aName="Unsupported protocol version")
```
akx, line 3352: This change doesn't seem strictly type-related - would be nice to be sure it's covered by tests, at least?

akx (changes requested):
A couple more comments - sorry it took a while for the second round of reviews. Starting to look pretty nice though, excellent renovation here :) 🛠️

src/paho/mqtt/client.py
PierreF: I've done this after adding type to the on_connect callback and saw that reason could be either a int or a ReasonCodes. I believe using int was a mistake.

Anyway I've added a test, and test show that it should no break compatibility (`reason == 132` work)

PierreF:
All your comments should be resolved. I think I'll squash-merge this PR due to multiple commits adding the removing thing. If you believe that we should kept multiple commits, I can do a rebase to only squash together part that belong to the same commit (but I want to squash or rebase, because their is change + exact revert of that change in this PR)

akx:
@PierreF A squash-merge, or a squashing rebase in this PR before merging sounds fine :) In fact, if #791 gets merged (and Codecov gets a baseline of the coverage for `master`), this could be squash-rebased on `master` and we'd see the coverage diff here too (hopefully).

PR #793, opened 2024-01-03
Add formating

PierreF:
Split of https://github.com/eclipse/paho.mqtt.python/pull/786 : contains add of formatting using ruff-format.

This PR will not be merged yet as it will need fewer pending PR to avoid conflict.


PierreF:
Too old PR and change auto-generated (by ruff format IIRC). New fresh PR should be re-created when auto-formatted will be introduced.

PR #794, opened 2024-01-06
Add properties to access Client attributes

### LINKED ISSUE #764: New properties for protected attributes so downstream can depend on them
I am hoping the PAHO team would be open to exposing some currently protected attributes as properties so downstream projects can / will couple to them.

Reference: https://github.com/sbtinstruments/aiomqtt/issues/191

Example of what is desired:

```python
class NewClient(mqtt.Client):
    """Paho client with some exposed protected values"""

    @property
    def client_id(self) -> str:
        return self._client_id.decode()

    @client_id.setter
    def client_id(self, value: str) -> None:
        if value == "":
            if self._protocol == ProtocolVersion.V31:
                self._client_id = base62(uuid4().int, padding=22)  # copied from paho code
            else:
                self._client_id = b""
        else:
            self._client_id = value.encode()

    @property
    def host(self) -> str:
        return self._host

    @host.setter
    def host(self, value: str):
        self._host = str(value)

    @property
    def port(self) -> int:
        return self._port

    @port.setter
    def port(self, value: int) -> None:
        value = int(value)
        if 0x0000 <= value <= 0xFFFF:
            self._port = value
        raise ValueError("port number out of range")

    @property
    def username(self) -> str | None:
        return None if self._username is None else self._username.decode()

    @username.setter
    def username(self, value: str | None):
        self._username = None if value is None else value.encode()

    @property
    def password(self) -> str | None:
        return None if self._password is None else self._password.decode()

    @password.setter
    def password(self, value: str | None):
        self._password = None if value is None else value.encode()

    @property
    def connect_timeout(self) -> float:
        return self._connect_timeout

    @connect_timeout.setter
    def connect_timeout(self, value: float) -> None:
        value = float(value)
        if value <= 0.0:
            raise ValueError("timeout must be a positive number")
        self._connect_timeout = value

    @property
    def keep_alive(self) -> int:
        return self._keepalive

    @keep_alive.setter
    def keep_alive(self, value: int) -> None:
        value = int(value)
        if value <= 0:
            raise ValueError("keepalive must be a positive number")
        self._keepalive = value

    @property
    def max_inflight_messages(self) -> int:
        return self._max_inflight_messages

    @max_inflight_messages.setter
    def max_inflight_messages(self, value: int) -> None:
        if value < 0:
            raise ValueError("value must not be negative")
        self._max_inflight_messages = value

    @property
    def max_queued_messages(self) -> int:
        return self._max_queued_messages

    @max_queued_messages.setter
    def max_queued_messages(self, value: int) -> None:
        if value < 0:
            raise ValueError("value must not be negative")
        self._max_queued_messages = value

    @property
    def will_topic(self) -> str:
        return self._will_topic.decode()

    @property
    def will_payload(self) -> bytes:
        return self._will_payload
```

PierreF:
Fix #764, add properties for
* host
* port
* keepalive
* transport
* protocol
* connect_timeout
* username
* password
* max_inflight_messages
* max_queued_messages
* will_topic
* will_payload
* logger


src/paho/mqtt/client.py
```diff
@@ -625,6 +624,196 @@ def __init__(self, client_id="", clean_session=None, userdata=None,
     def __del__(self):
         self._reset_sockets()
 
+    @property
+    def host(self) -> str:
+        "Host we try to connect to. If don't yet called connect() return empty string"
```
akx, line 629:
```suggestion
        """Host to connect to. If `connect()` hasn't been called yet, returns an empty string."""
```

src/paho/mqtt/client.py
```diff
@@ -625,6 +624,196 @@ def __init__(self, client_id="", clean_session=None, userdata=None,
...
+        Update host. This will only be used on future (re)connection. You should probably
+        use reconnect() to update the connection if established.
+        """
+        if value is None or len(value) == 0:
+            raise ValueError("Invalid host.")
+        self._host = value
+
+    @property
+    def port(self) -> int:
+        "TCP port we try to connect to."
```
akx, line 644:
```suggestion
        """Broker TCP port to connect to."""
```

src/paho/mqtt/client.py
```diff
@@ -625,6 +624,196 @@ def __init__(self, client_id="", clean_session=None, userdata=None,
...
+        return self._port
+
+    @port.setter
+    def port(self, value: int) -> None:
+        """
+        Update post. This will only be used on future (re)connection. You should probably
```
akx, line 650:
```suggestion
        Update port. This will only be used on future (re)connection. You should probably
```

src/paho/mqtt/client.py
```diff
@@ -625,6 +624,196 @@ def __init__(self, client_id="", clean_session=None, userdata=None,
...
+        use reconnect() to update the connection if established.
+        """
+        if value <= 0:
+            raise ValueError("Invalid port number.")
+        self._port = value
+
+    @property
+    def keepalive(self) -> int:
+        "Keepalive in seconds used by the client "
```
akx, line 659:
```suggestion
        """Client keepalive interval (in seconds)."""
```

src/paho/mqtt/client.py
```diff
@@ -625,6 +624,196 @@ def __init__(self, client_id="", clean_session=None, userdata=None,
...
+    def keepalive(self, value: int) -> None:
+        "Update keepalive. It's behavior is undefined if the connection is already open"
+        if self._sock is not None:
+            # The issue here is that the previous value of keepalive matter to possibly
+            # sent ping packet.
+            warnings.warn(
+                "updating keepalive on established connection is not supported",
+                stacklevel=2,
+            )
+            self._sock.settimeout(value)
```
akx, line 672: I think this should just raise a RuntimeError if the behavior is undefined?

src/paho/mqtt/client.py
```diff
@@ -625,6 +624,196 @@ def __init__(self, client_id="", clean_session=None, userdata=None,
...
+        self._port = value
+
+    @property
+    def keepalive(self) -> int:
+        "Keepalive in seconds used by the client "
+        return self._keepalive
+
+    @keepalive.setter
+    def keepalive(self, value: int) -> None:
+        "Update keepalive. It's behavior is undefined if the connection is already open"
```
akx, line 664:
```suggestion
        """Update the client keepalive interval. This may not be called if the connection is already open."""
```

src/paho/mqtt/client.py
```diff
@@ -625,6 +624,196 @@ def __init__(self, client_id="", clean_session=None, userdata=None,
...
+            self._sock.settimeout(value)
+
+        if value < 0:
+            raise ValueError("Keepalive must be >=0.")
+
+        self._keepalive = value
+
+    @property
+    def transport(self) -> str:
+        'Transport used for the connection, could be "tcp" or  "websockets".'
```
akx, line 681: I'd rather skip the "could be" since we don't really mandate it.
```suggestion
        """Transport method used for the connection."""
```

src/paho/mqtt/client.py
```diff
@@ -625,6 +624,196 @@ def __init__(self, client_id="", clean_session=None, userdata=None,
...
+        """
+        if value.lower() not in ("websockets", "tcp"):
+            raise ValueError(
+                f'transport must be "websockets" or "tcp", not {value}')
+
+        self._transport = value.lower()
+
+    @property
+    def protocol(self) -> int:
+        "Protocol version used (MQTT v3, MQTT v3.11, MQTTv5)"
```
akx, line 699: This doesn't specify what the returned values will be, but I suspect this PR will only be merged after #791, which adds an enumeration for the protocol version?
```suggestion
        """Protocol version used (MQTT v3, MQTT v3.11, MQTTv5)"""
```

src/paho/mqtt/client.py
```diff
@@ -625,6 +624,196 @@ def __init__(self, client_id="", clean_session=None, userdata=None,
...
+        return self.protocol
+
+    @property
+    def connect_timeout(self) -> float:
+        "Timeout used to establish the TCP (& TLS / websocket if enabled) in seconds"
```
akx, line 704:
```suggestion
        """Connection establishment timeout in seconds"""
```

src/paho/mqtt/client.py
```diff
@@ -625,6 +624,196 @@ def __init__(self, client_id="", clean_session=None, userdata=None,
...
+    def connect_timeout(self, value: float):
+        "Change connect_timeout for future (re)connection"
+        if value <= 0.0:
+            raise ValueError("timeout must be a positive number")
+
+        self._connect_timeout = value
+
+    @property
+    def username(self) -> str | None:
+        "Return the username use to connect to MQTT broken or None if not credenials are used."
```
akx, line 717:
```suggestion
        """The username used to connect to the MQTT broker, or None if no username is used."""
```

src/paho/mqtt/client.py
```diff
@@ -625,6 +624,196 @@ def __init__(self, client_id="", clean_session=None, userdata=None,
...
+        use reconnect() to update the connection if established.
+        """
+        if value is None:
+            self._username = None
+        else:
+            self._username = value.encode("utf-8")
+
+    @property
+    def password(self) -> str | None:
+        "Return the password use to connect to MQTT broken or None if not password are used."
```
akx, line 735:
```suggestion
        """The password used to connect to the MQTT broker, or None if no password is used."""
```

src/paho/mqtt/client.py
```diff
@@ -625,6 +624,196 @@ def __init__(self, client_id="", clean_session=None, userdata=None,
...
+        use reconnect() to update the connection if established.
+        """
+        if value is None:
+            self._password = None
+        else:
+            self._password = value.encode("utf-8")
+
+    @property
+    def max_inflight_messages(self) -> int:
+        "Maximum number of message with QoS > 0 that can be part way through their network flow at once"
```
akx, line 753:
```suggestion
        """Maximum number of messages with QoS > 0 that can be partway through the network flow at once"""
```
... but what does "partway through the network flow" mean? It's not quite clear.

src/paho/mqtt/client.py
```diff
@@ -625,6 +624,196 @@ def __init__(self, client_id="", clean_session=None, userdata=None,
...
+            )
+
+        if value < 0:
+            raise ValueError("Invalid inflight.")
+
+        self._max_inflight_messages = value
+
+    @property
+    def max_queued_messages(self) -> int:
+        "Maximum number of message with QoS > 0 that can be part way through their network flow at once"
```
akx, line 774:
```suggestion
        """Maximum number of messages with QoS > 0 that can be partway through the network flow at once"""
```
... same concern about "partway through the network flow" – and then, this string is exactly the same as `max_inflight_messages`, so I'm not sure what the actual difference is? I should hope there is a difference.

src/paho/mqtt/client.py
```diff
@@ -625,6 +624,196 @@ def __init__(self, client_id="", clean_session=None, userdata=None,
...
+
+    @max_queued_messages.setter
+    def max_queued_messages(self, value: int) -> None:
+        "Update max_queued_messages. It's behavior is undefined if the connection is already open"
+        if self._sock is not None:
+            # Not tested.
+            warnings.warn(
+                "updating max_queued_messages on established connection is not supported",
+                stacklevel=2,
+            )
```
akx, line 785: Should just raise a RuntimeError IMO.

src/paho/mqtt/client.py
```diff
@@ -625,6 +624,196 @@ def __init__(self, client_id="", clean_session=None, userdata=None,
...
+    @max_inflight_messages.setter
+    def max_inflight_messages(self, value: int) -> None:
+        "Update max_inflight_messages. It's behavior is undefined if the connection is already open"
+        if self._sock is not None:
+            # Not tested. Some doubt that everything is okay when max_inflight change between 0
+            # and > 0 value because _update_inflight is skipped when _max_inflight_messages == 0
+            warnings.warn(
+                "updating max_inflight_messages on established connection is not supported",
+                stacklevel=2,
+            )
```
akx, line 765: Should just raise a RuntimeError.

akx (changes requested):
Grammatical etc. fixes within – and also, I'm not sure there should be setters at all for the properties which won't take an effect until reconnect? Are there actual use cases for those?

src/paho/mqtt/client.py
PierreF: Here it's a copy/paste mistake I made. Should the same description of max_queued_messages_set which don't imply network, only internal queue.

PierreF:
I've added setter, because some properties can't be changed without setter (at least connection_timeout), and then I've added setter for other for consistency. But I'm also unsure this will be very useful.

src/paho/mqtt/client.py
```diff
@@ -625,6 +624,196 @@ def __init__(self, client_id="", clean_session=None, userdata=None,
...
+
+        return self._will_topic.decode("utf-8")
+
+    @property
+    def will_payload(self) -> bytes | None:
+        "Return the payload will send on unexpected disconnect, or None if will isn't set"
+        if self._will_topic is None:
+            return None
+
+        return self._will_topic.decode("utf-8")
```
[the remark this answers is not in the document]
PierreF: It's also odd that we use `self._will_TOPIC` in the will_PAYLOAD property. Copy/paste is not always your friend :)

PierreF:
I'm going with disallowing (raise RuntimeError) changing properties that need to reconnect. It's only allowed to change them when not connected or *connecting*. It still disallowed if the connection is currently down because a loop_forever or loop_start might reconnect concurrently. It's required to disconnect() which ensure no automatic reconnection occur

PR #795, opened 2024-01-07
Fix is_connected property when not using loop_forever

### LINKED ISSUE #525: State of client doesn't change to disconnected after connection lost
When not using loop_forever or loop_start and running the loop in the main thread, a hard disconnect gets captured by the client but doesn't change the output of `is_connected()` method.

Steps to reproduce:

1. Run an mqtt server, I use emqx:

```bash    
docker run -d --name emqx -p 18083:18083 -p 1883:1883 emqx/emqx:v4.0.0
```

2. Run the code below:
   
```python
import time
import logging

logging.basicConfig(level=logging.DEBUG)

logger = logging.getLogger(__name__)


    logger.debug('Connected with result code ' + str(rc))
    # subscribe
    client.subscribe('tienda/#')


def on_message(client, userdata, msg):
    logger.debug(msg.topic + " " + str(msg.payload))
    # if b'stop' in msg.payload:
    #     client.disconnect()


def on_disconnect(client, flags, rc):
    logger.debug('Disconnected with result code ' + str(rc))
    # client.reconnect()

# Build the client

client = mqtt.Client(clean_session=True)
logger.error('Client initialized with status: %s', client._state)
# Specify callback function
client.on_message = on_message
client.on_disconnect = on_disconnect
client.enable_logger(logger)
# client.username_pw_set('topse', 'cret')


# Establish a connection
client.connect('localhost', 1883, 60)
# wait for connection
while not client.is_connected():
    logger.debug('Connecting')
    client.loop(timeout=1)

logger.debug('Connected')

# This loops never ends even after there is no connection at all
while client.is_connected():
    client.loop()
    time.sleep(1)
    logger.debug('In main loop: %s', client.is_connected())
```

3. browse http://localhost:18083/#/clients and find you client and disconnect it.

The result is that you will see the `Disconnected with result code: 0` message, but the program never exits and keeps printing:

```bash
DEBUG:__main__:In main loop: True
DEBUG:__main__:In main loop: True
DEBUG:__main__:In main loop: True
DEBUG:__main__:In main loop: True
DEBUG:__main__:In main loop: True
```

The problem is the internal field `_state` doesn't changes to disconnected or any other states when we are not calling the disconnect method directly. This causes the `is_connected()` method return true and the loop never exits.

paho-mqtt version: 1.5.1

PierreF:
Fix #525

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -1778,7 +1778,6 @@ class Client:
                 rc = MQTTErrorCode.MQTT_ERR_SUCCESS
             else:
                 self._state = ConnectionState.MQTT_CS_CONNECTION_LOST
-                self._easy_log(MQTT_LOG_DEBUG, "... 2")
                 rc = MQTTErrorCode.MQTT_ERR_KEEPALIVE
 
             self._do_on_disconnect(rc)
@@ -2592,7 +2591,6 @@ class Client:
             self._do_on_disconnect(rc, properties)
 
         if rc == MQTT_ERR_CONN_LOST:
-            self._easy_log(MQTT_LOG_DEBUG, "... 3")
             self._state = ConnectionState.MQTT_CS_CONNECTION_LOST
 
         return rc
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -1290,10 +1290,19 @@ class Client:
             socklist = select.select(rlist, wlist, [], timeout)
         except TypeError:
             # Socket isn't correct type, in likelihood connection is lost
+            # ... or we called disconnect(). In that case the socket will
+            # be closed but some loop (like loop_forever) will continue to
+            # call _loop(). We still want to break that loop by returning an
+            # rc != MQTT_ERR_SUCCESS and we don't want state to change from
+            # mqtt_cs_disconnecting.
+            if self._state != ConnectionState.MQTT_CS_DISCONNECTING:
+                self._state = ConnectionState.MQTT_CS_CONNECTION_LOST
             return MQTTErrorCode.MQTT_ERR_CONN_LOST
         except ValueError:
             # Can occur if we just reconnected but rlist/wlist contain a -1 for
             # some reason.
+            if self._state != ConnectionState.MQTT_CS_DISCONNECTING:
+                self._state = ConnectionState.MQTT_CS_CONNECTION_LOST
             return MQTTErrorCode.MQTT_ERR_CONN_LOST
         except Exception:
             # Note that KeyboardInterrupt, etc. can still terminate since they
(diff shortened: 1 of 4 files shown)


src/paho/mqtt/client.py
```diff
@@ -600,7 +601,8 @@ def __init__(self, client_id="", clean_session=None, userdata=None,
         self._ssl_context = None
         # Only used when SSL context does not have check_hostname attribute
         self._tls_insecure = False
-        self._logger = None
+        logging.basicConfig(level=logging.DEBUG)
+        self._logger = logging.getLogger("plop")
```
akx, line 605: plop! 😂

akx (commented):
I think there are stray logger initializations and log calls in here :)

A test would probably not go amiss either.

src/paho/mqtt/client.py
PierreF: Thanks, totally missed this temporary debugging code.

src/paho/mqtt/client.py
```diff
@@ -1762,6 +1771,8 @@ def loop_misc(self) -> MQTTErrorCode:
             if self._state == mqtt_cs_disconnecting:
                 rc = MQTTErrorCode.MQTT_ERR_SUCCESS
             else:
+                self._state = ConnectionState.MQTT_CS_CONNECTION_LOST
+                self._easy_log(MQTT_LOG_DEBUG, "... 2")
```
akx, line 1775:
```suggestion
```

src/paho/mqtt/client.py
```diff
@@ -2574,6 +2585,10 @@ def _loop_rc_handle(
 
             self._do_on_disconnect(rc, properties)
 
+        if rc == MQTT_ERR_CONN_LOST:
+            self._easy_log(MQTT_LOG_DEBUG, "... 3")
```
akx, line 2589:
```suggestion
```

tests/test_client.py
```diff
@@ -137,6 +139,116 @@ def on_connect(mqttc, obj, flags, reason, properties):
...
+            fake_broker.finish()
+
+            # call loop() to detect the connection lost
+            assert mqttc.loop(timeout=1) == MQTTErrorCode.MQTT_ERR_CONN_LOST
+
+            assert on_disconnect_reached.wait(1)
+            assert not mqttc.is_connected()
+
+        finally:
+            mqttc.loop_stop()
```
akx, line 250: I don't see `loop_start` being called in this test..?

akx (changes requested):
There's a couple more debug log calls remaining, as well as what I think may be a stray `loop_stop()` (in a more perfect world, we'd use e.g. py.test fixtures or a `with` block to manage `loop_stop()` always getting called, but that's out of scope for this PR)

src/paho/mqtt/client.py
PierreF: My bad, I should re-read my own PR more carefully :(

PR #796, opened 2024-01-07
Fix wait_for_publish that could hang for QoS=0 message

### LINKED ISSUE #549: loop_start() before connect()
The documentation says that loop_start() can be called either before or after connect(), however this appears not to be true. Calling loop_start() after connect() seems to work fine

    import paho.mqtt.client
    print(paho.mqtt.__version__)

    client = paho.mqtt.client.Client()

    client.connect('keylimepi.lan')
    client.loop_start()

    mi = client.publish('test', 'testing')
    mi.wait_for_publish()

    client.disconnect()
    client.loop_stop()

but if I call it before

    import paho.mqtt.client
    print(paho.mqtt.__version__)

    client = paho.mqtt.client.Client()

    client.loop_start()
    client.connect('keylimepi.lan')

    mi = client.publish('test', 'testing')
    mi.wait_for_publish()

    client.disconnect()
    client.loop_stop()

then the message  is not published and the wait_for_publish() call hangs.

This behaviour is observed with both version 1.4.0 and 1.5.1

PierreF:
If reconnect() is called (which happen on auto-reconnection or on first connection done by loop_forever), any pending QoS = 0 message were dropped without unblocking wait_for_publish().

We now ensure wait_for_publish() is unblocked and raise RuntimeError with MQTT_ERR_CONN_LOST

Fix #549

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -455,7 +455,7 @@ class MQTTMessageInfo:
         with self._condition:
             while not self._published and not timed_out():
                 self._condition.wait(timeout_tenth)
-        
+
         if self.rc > 0:
             raise RuntimeError(f'Message publish failed: {error_string(self.rc)}')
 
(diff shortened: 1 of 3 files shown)


src/paho/mqtt/client.py
```diff
@@ -349,6 +349,10 @@ def timed_out():
             while not self._published and not timed_out():
                 self._condition.wait(timeout_tenth)
 
+        if self.rc > 0:
+            raise RuntimeError(f'Message publish failed: {error_string(self.rc)}')
+
+
```
akx, line 355:
```suggestion
```

src/paho/mqtt/client.py
```diff
@@ -1018,8 +1022,16 @@ def reconnect(self):
...
 
+        # Before dropping all out_packet, ensure any QoS == 0 message info get
+        # marked as MQTT_ERR_CONN_LOST or the wait_for_publish() could hang forever
+        old_queue = self._out_packet
         self._out_packet = collections.deque()
 
+        for pkt in old_queue:
+            if pkt["command"] & 0xF0 == PUBLISH and pkt["qos"] == 0:
+                pkt["info"].rc = MQTT_ERR_CONN_LOST
+                pkt["info"]._set_as_published()
```
akx, line 1033: Is there some subtlety here as to why `old_queue` is grabbed first and this is only done exactly after `self._out_packet` has been reallocated? If not, it feels like

```suggestion
        # Mark all currently outgoing QoS = 0 packets as lost,
        # or `wait_for_publish()` could hang forever
        for pkt in self._out_packet:
            if pkt["command"] & 0xF0 == PUBLISH and pkt["qos"] == 0:
                pkt["info"].rc = MQTT_ERR_CONN_LOST
                pkt["info"]._set_as_published()
        
        self._out_packet = collections.deque()
```
would be the more obvious way to write this.

(`self._out_packet.clear()` would also save some allocations.)

akx (commented):
Should there be a test for this?

src/paho/mqtt/client.py
PierreF: The reason way for what happen if a `publish()` if call during `reconnect()`. It would end in the new `_out_packet` or be marked at published... but well it's not very nice.

I'll check, but I think it better in all aspect to close the socket, then process the _set_as_published + clear the out_packet queue, because once closed we don't add message to out_packet.

PR #797, opened 2024-01-07
Correctly mark connection as broken on SSL error and don't crash loop_forever

### LINKED ISSUE #750: SSL EOF leads to persistent non-delivery of messages
I am using 1.6.1, NetBSD 9 amd64, python 3.10.  The broker is tup-to-date mosquitto, on a remote NetBSD 9 amd64 system.  I am using 8883 with a real certificate, and generally everything works well.

I have a script which polls a UPS and publishes messages usually once a minute, but at intervals of up to 1s if something interesting has happened.  On that system, I took the WAN interface down, releasing the lease, removing addresses and default route, and `ifconfig down`, waited an hour and brought it back.   Obviously I won't have data in the meantime but I expect it to recover.

In the log (stdout/stderr) of the program I see (with the json shortened; there are actually 14 keys):

```
SUBMITTED time {'time': 1696117812.470824, 'runtime': 5940.0}
PUBLISHED 16075
SUBMITTED time {'time': 1696117873.4957478, 'runtime': 5940.0}
Exception in thread Thread-1 (_thread_main):
  File "/usr/pkg/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/usr/pkg/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/pkg/lib/python3.10/site-packages/paho/mqtt/client.py", line 3591, in _thread_main
    self.loop_forever(retry_first_connection=True)
  File "/usr/pkg/lib/python3.10/site-packages/paho/mqtt/client.py", line 1756, in loop_forever
    rc = self._loop(timeout)
  File "/usr/pkg/lib/python3.10/site-packages/paho/mqtt/client.py", line 1181, in _loop
    rc = self.loop_write()
  File "/usr/pkg/lib/python3.10/site-packages/paho/mqtt/client.py", line 1577, in loop_write
    rc = self._packet_write()
  File "/usr/pkg/lib/python3.10/site-packages/paho/mqtt/client.py", line 2464, in _packet_write
    write_length = self._sock_send(
  File "/usr/pkg/lib/python3.10/site-packages/paho/mqtt/client.py", line 649, in _sock_send
    return self._sock.send(buf)
  File "/usr/pkg/lib/python3.10/ssl.py", line 1206, in send
    return self._sslobj.write(data)
ssl.SSLEOFError: EOF occurred in violation of protocol (_ssl.c:2426)
PUBLISHED 16076
SUBMITTED time {'time': 1696117934.524036, 'runtime': 5940.0}
SUBMITTED time {'time': 1696117994.8527699, 'runtime': 5940.0}
```

So the "EOF" looks like write failed.  Probably this is "no route to host".    However I don't get a disconnect callback.   And I did get a publish callback!  But surely the broker didn't get the message.
Then, future publish calls happen, with no exceptions.  But no publish callbacks.

Obviously I can write a watchdog to close/open if I don't get pubacks to python.

But it seems that the ssl error should be caught and the connection should be judged non-functional.   Or at least some later publish should cause failure and to close the connection.  The user program should not get an exception thrown other than in response to an API call and then only documented exceptions.

I don't think this is related to threading/locking, as the program only did one publish call, and without the network being down, runs for months.

PierreF:
Catch all OSError as indicator of connection broken. It should especially include SSL error. Also make WebsocketConnectionError a subclass of ConnectionError (which is a subclass of OSError).

Should fix #750.

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -27,6 +27,8 @@ v2.0.0 - 2023-xx-xx
   and loop_start/loop_forever isn't used. Closes #525.
 - Fix wait_for_publish that could hang with QoS == 0 message on reconnection
   or publish during connection. Closes #549.
+- Correctly mark connection as broken on SSL error and don't crash loop_forever.
+  Closes #750.
 
 
 v1.6.1 - 2021-10-21
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -239,7 +239,7 @@ CallbackOnUnsubscribe = Union[
 _socket = socket
 
 
-class WebsocketConnectionError(ValueError):
+class WebsocketConnectionError(ConnectionError):
     pass
 
 
@@ -1909,7 +1909,7 @@ class Client:
             if self._state == mqtt_cs_connect_async:
                 try:
                     self.reconnect()
-                except (OSError, WebsocketConnectionError):
+                except OSError:
                     self._handle_on_connect_fail()
                     if not retry_first_connection:
                         raise
@@ -1950,7 +1950,7 @@ class Client:
                 else:
                     try:
                         self.reconnect()
-                    except (OSError, WebsocketConnectionError):
+                    except OSError:
                         self._handle_on_connect_fail()
                         self._easy_log(
                             MQTT_LOG_DEBUG, "Connection failed, retrying")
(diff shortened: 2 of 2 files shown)


gdt:
I've been running this a few days including a few netflakes and my client is connected.  I can't prove it is right but I lean to merging it.

PierreF:
I've made a revert on Websocket wrapper connection due to issue catched during additional tests (https://github.com/eclipse/paho.mqtt.python/pull/797/commits/f0e63d2322190d8526f20ef0c0928a88a0898943).

PR #804, opened 2024-01-21
Update readme and example

PierreF:
This update the README and examples:

* To fix breaking change in example present inside the READ
* To fix breaking change in examples directory
* It drop duplicate documentation in README that is also present as docstring. Have duplicated documentation is too much work with no gain.
  I want to kept README being a guide / example of usage. And docstring is the reference of all functions with all arguments.

I've added Sphinx documentation (only expose docstring for now). There is a markdown export to serve them from the Github repository, but that more a workaround because finding a right place to host the documentation (readthedocs ?)


PierreF:
The PR is based of #808 for conflict but also because I spotted those private objects while checking the documentation output.

PierreF:
The documentation is now available online at https://eclipse.dev/paho/files/paho.mqtt.python/html/index.html

README.rst
```diff
@@ -110,9 +107,9 @@ You should set ``clean_session = False`` if you need the QoS 2 guarantee of only
 Usage and API
 -------------
 
-Detailed API documentation is available through **pydoc**. Samples are available in the **examples** directory.
+Detailed API documentation `is available online <documentation_>`_ or could be build from ``docs/`` and samples are available in the `examples`_ directory.
```
dannystaple, line 110:
```suggestion
Detailed API documentation `is available online <documentation_>`_ or could be built from ``docs/`` and samples are available in the `examples`_ directory.
```

PR #806, opened 2024-01-23
Nicer print in Getting Started example

laundmo:
Using the connack_string and error_string functions to print nicer messages in the Getting Started example.

Makes it easier for users to understand issues without searching for lists of these codes online

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -125,7 +125,7 @@ Here is a very simple example that subscribes to the broker $SYS topic tree and 
 
     # The callback for when the client receives a CONNACK response from the server.
     def on_connect(client, userdata, flags, rc):
-        print("Connected with result code "+str(rc))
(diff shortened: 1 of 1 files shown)


PierreF:
I believe this will no longer be needed with 2.0 version that is arriving soon (a pre-release is already available). With the newer version, when using the CallbackAPI version 2, it's no longer an integer `rc` that is received but an ReasonCodes instance, which is printed as string.

Not everything is well aligned currently: master branch contains the development branch with latest change (the 2.0 version) but the README is not yet updated. See https://github.com/eclipse/paho.mqtt.python/pull/804 for README matching version 2.0

So the getting started will result in:
```
>>>  print(f"Connected with result code {reason_code}")
Connected with result code Success
```

PierreF:
Closing as I believe the issue is fixed with latest release. Feel free to re-open if you still have an issue.

PR #809, opened 2024-02-02
Fix loop_stop() not clearing ._thread when called from same thread

jiachin1995:
When client.loop_stop() is called from within on_message() callback, client ._thread will not be set to None. This will cause subsequent loop_start() to return MQTT_ERR_INVAL error.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -2274,7 +2274,7 @@ class Client:
         self._thread_terminate = True
         if threading.current_thread() != self._thread:
             self._thread.join()
-            self._thread = None
+        self._thread = None
 
         return MQTTErrorCode.MQTT_ERR_SUCCESS
 
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -2274,7 +2274,6 @@ class Client:
         self._thread_terminate = True
         if threading.current_thread() != self._thread:
             self._thread.join()
-        self._thread = None
 
         return MQTTErrorCode.MQTT_ERR_SUCCESS
 
@@ -4420,6 +4419,7 @@ class Client:
 
     def _thread_main(self) -> None:
         self.loop_forever(retry_first_connection=True)
+        self._thread = None
 
     def _reconnect_wait(self) -> None:
         # See reconnect_delay_set for details
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -4418,8 +4418,10 @@ class Client:
                         MQTT_LOG_ERR, 'Caught exception in on_connect_fail: %s', err)
 
     def _thread_main(self) -> None:
-        self.loop_forever(retry_first_connection=True)
-        self._thread = None
+        try:
+            self.loop_forever(retry_first_connection=True)
+        finally:
+            self._thread = None
 
     def _reconnect_wait(self) -> None:
         # See reconnect_delay_set for details


jiachin1995:
Context:
  Our use case was for PahoMQTT to simulate an IoT device.  The device can be remotely rebooted/shutdown.
  If we try to call loop_stop from within the PahoMQTT thread. It successfully stops the thread but subsequently loop_start() stops working as client._thread is not None.

PierreF:
Hello, thanks for your contribution. For a contribution to be merged, you will need to sign the Eclipse ECA for your contribution to be merged. (https://accounts.eclipse.org/user/eca)

That being said, if I understand correctly your use-case, you do not want to (just) call loop_stop/loop_start. This will mostly do nothing, it do not disconnect from MQTT (so it don't simulate a device reboot).
loop_stop only stop the network processing thread, and if you call loop_start quickly, you will only produce a short delay in network processing.

I'm not sure if could be valid to call loop_stop & loop_start within a callback. Because to call loop_start I believe it's required to have the thread stopped (or else it will be too complex regarding possible race-condition), and the thread won't stop until the callback returned.

Actually if you want to simulate de device reboot, you should also discard the Client state (like the possibly queued message), for that the best solution would be to re-create the Client(). For this the best solution IMO is doing something like:
```

def on_message(...):
   if should_simulate_device_reboot():
      # calling disconnect() will *cleanly* disconnect - so it don't simulate a device lost, but device clean reboot
      # but calling disconnect() will stop the loop_forever()
      client.disconnect() 

while simulator_should_run:
   # we re-create the mqtt.Client() each time loop_forever exit
   mqttc = mqtt.Client([...])
   mqttc.connect[...]
   mqttc.loop_forever()
```

Does this solve your problem ?

jiachin1995:
Hi Pierre, I have signed the ECA.

For restart, I have already assessed and agree that there is no meaningful way to implement it in PahoMQTT. I simply wanted to provide some context for the error we faced. We are fine with creating a separate thread.

PierreF:
Thank for your patience. I do have fear that your solution will introduce new bugs, mostly because the thread will still be running with `self._thread == None`. This might break some assumption in current code.

If I get correctly the aim of this fix (being able to call `loop_stop()` from `on_message()` and later calling `loop_start()`), shouldn't we prefer to set `self._thread = None` in a try/finally of `_thread_main()` ? This should fix the issue AND ensure the thread isn't running with `self._thread == None` ?

This should also make the "later" to call `loop_start()` possible to determine. It's when `loop_start()` don't fail with `MQTT_ERR_INVAL`. With your proposition I don't see a way to be sure that isn't two threads running.

jiachin1995:
Hi Pierre. I have updated the code as suggested. However, I am unsure where to insert a try/finally block. 

Do let me know if any other changes are required

src/paho/mqtt/client.py
```diff
@@ -4420,6 +4419,7 @@ def _handle_on_connect_fail(self) -> None:
 
     def _thread_main(self) -> None:
         self.loop_forever(retry_first_connection=True)
+        self._thread = None
```
PierreF, line 4422: I was thinking to wrap call to loop_forever in _thread_main in a try/finally to be sure `_thread` is set to None even in case of error:

```suggestion
        try:
            self.loop_forever(retry_first_connection=True)
        finally:        
            self._thread = None
```

This only makes change in case loop_forever raise an error.

jiachin1995: HI Pierre. Thanks for the explanation.

I have made the suggested changes.

Issue #789, opened 2024-01-03
Tests are flaky on `master`


akx:
Tests are currently flaky on `master`. These seem to be the newly re-enabled MQTTv5 tests.

@PierreF Instead of `paho.mqtt.testing`, would it make sense to just use Mosquitto here?

PierreF:
MQTTv5 test aren't optimal in current form. They are very slow (I've quickly looked and didn't find a way to make them run faster easily).

Yes doing test using a real broker like mosquitto might be a good solution.

PierreF:
Done some fix in e96ff7578f7f56a7f3068fbc3e7d0b48e1c64a97 which seems good enough. The last run on master worker without need to retry.

PR #821, opened 2024-02-29
Return correct object as client protocol

### LINKED ISSUE #820: Reading Client protocol property causes infinite recursion loop
# Bug Description

Accessing the field `Client.protocol` causes infinite recursion, since there is a bug in how that property is defined. The bug is on [this line](https://github.com/eclipse/paho.mqtt.python/blob/613520775c5706f72b916a99a8931983633fe489/src/paho/mqtt/client.py#L956). The function should return `self._protocol` and not `self.protocol` as the latter is a recursive definition.
This is the property.
```python
    @property
    def protocol(self) -> MQTTProtocolVersion:
        """
        Protocol version used (MQTT v3, MQTT v3.11, MQTTv5)

        This property is read-only.
        """
        return self.protocol
```

asikeero:
Fixes #820

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -953,7 +953,7 @@ class Client:
 
         This property is read-only.
         """
-        return self.protocol
+        return self._protocol
 
     @property
     def connect_timeout(self) -> float:


PierreF:
Thanks for catching this error. Could you sign ECA (https://accounts.eclipse.org/user/eca / https://www.eclipse.org/legal/ECA.php) so I can merge you PR ?

asikeero:
Done, should be OK now.

asikeero:
Two checks are failing due to reasons that seem not to relate to this PR. Just to clarify, am I expected to add something to this PR to make them pass?

PierreF:
Thanks for your patience. Pytest (tox) tend to be have false-positive errors (cf #789). The lint_python error is a real error, linked to your change.
Your change isn't the issue actually, but make the issue visible: `self._protocol` is not correctly typed.

I'll merge your PR and fix the typing once merged.

Thanks for your contribution.

PR #825, opened 2024-03-11
remove superfluous lock

smurfix:
Assigning to a variable is atomic anyway.
The lock is thus superfluous.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -836,7 +836,6 @@ class Client:
         self._bind_port = 0
         self._proxy: Any = {}
         self._in_callback_mutex = threading.Lock()
-        self._callback_mutex = threading.RLock()
         self._msgtime_mutex = threading.Lock()
         self._out_message_mutex = threading.RLock()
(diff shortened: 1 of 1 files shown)


src/paho/mqtt/client.py
```diff
@@ -3001,8 +2983,7 @@ def handle_mytopic(client, userdata, message):
         if callback is None or sub is None:
             raise ValueError("sub and callback must both be defined.")
 
-        with self._callback_mutex:
-            self._on_message_filtered[sub] = callback
+        self._on_message_filtered[sub] = callback
```
PierreF, line 2986: This isn't a simple assignment. I'm unsure whether this is safe to remove the lock.

PierreF (commented):
What's the issue we try to solve by removing the locks ?

To my knowledge, this is safe to remove on CPython due to GIL and that Python *bytecode* are atomic. I don't known if this fact is documented by Python / CPython. I'm not sure this hold for other Python implementation (PyPy, Jython...) and I'm also unsure Python recommend relying on bytecode being atomic.

PR #829, opened 2024-03-25
add support for "unix" transport where socket module contains AF_UNIX

akrpic77:
This pull request adds "unix" transport  where underlying python's socket module contains AF_UNIX constant. This is, of course, available on all the unices, and since 2017 on Windows 10 too. Although support for it has [not yet ](/python/cpython/issues/77589) landed in the cpython.

The use case for this transport is when both broker and client run on the same machine - unix sockets should provide [noticable](https://scottmoonen.com/2008/04/05/a-performance-comparison-of-af_unix-with-loopback-on-linux/) performance gains.

Few remarks:
1) unix socket is created instead of normal socket, so it is possible to create tls over unix socket, but that is not tested
2) for client tests, the fake_broker is parametrized for 'tcp', and 'unix', this could also be used for e.g. 'tcp6' test
3) unix socket filename is currently 'localhost', so it wasn't necessary to modify connect* methods. I don't like it, maybe fake_broker.addr can be added as a tuple with value  (host, port) for "tcp" , and (path,) for "unix". This addr field would also come handy later if "tcp6" tests are added.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -682,6 +682,10 @@ class Client:
 
     :param transport: use "websockets" to use WebSockets as the transport
         mechanism. Set to "tcp" to use raw TCP, which is the default.
+        Use "unix" to use Unix sockets as the transport mechanism; note that
+        this option is only available on platforms that support Unix sockets,
+        and the "host" argument is interpreted as the path to the Unix socket
+        file in this case.
 
     :param bool manual_ack: normally, when a message is received, the library automatically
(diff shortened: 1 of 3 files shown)


PierreF:
Hello, thanks for your contribution. Could you sign the Eclipse ECA for your contribution to be merged (https://accounts.eclipse.org/user/eca) ?

akrpic77:
Done!

PierreF (approved):

PR #843, opened 2024-05-18
Add missing mapping for setsockopt to _WebsocketWrapper

bdraco:
Fixes the asyncio example with a websocket wrapped socket https://github.com/eclipse/paho.mqtt.python/blob/d45de3737879cfe7a6acc361631fa5cb1ef584bb/examples/loop_asyncio.py#L92


bdraco:
Tried to sign the CLA but I keep getting a 401 error.  Will try later

bdraco:
Seems to be ok now

PR #845, opened 2024-05-22
"See ReasonCode.names": did not work

m-u-xyz:
As this class attribute is actually useful,
make the code fit the comment.

This also saves a bit of time and memory because the map was recreated for each ReasonCode instance.


smurfix:
I added hopefully-uncontroversial fixes for a couple of other issues.

pswsm:
This PR can be closed, the issue was addressed on #913

Issue #42, opened 2016-03-18
keepalive value of 0 causes client to enter into indefinite loop


akashgoswami:
In one of the test cases, keepalive option was passed to 0 
This caused MQTT client to enter into a never ending loop and it keept on creating connection with broker and destroying them. Client was stuck at following location:

```
client.py", line 1281, in loop_forever
    time.sleep(1)

```
The behavior was same with **client.connect()** functions as well as **publish** functions.

Ideally keepalive of 0 should either cause an invalid argument return or should internally revert to a default value.

ralight:
Thanks, you're quite right. keepalive=0 is a perfectly valid setting, this code now just accepts it properly.

akashgoswami:
Awesome! Thanks for super fast response and fix.

PR #876, opened 2024-12-18
Seperate keepalive 0 from socket `settimeout` 

cjavad:
In certain conditions (MQTT over SSL, and also some websockets stuff) _keepalive is used as a socket timeout, which is all good until used with `0`, which triggers as #42 describes it "an infinite loop" preventing any connection from being established. Originally in our testing only the SSL code path was triggered, and referring an older change (#578) _connect_timeout can be used as an alternative, which also semantically makes sense. 

I noticed the same value was used on another socket for websocket transports, and with some quick testing both cases fail with keepalive 0:

Using test.mosquitto.org with websockets:

```python
...
>>> client.connect(url, port, keepalive=0)
... (stack trace)
BlockingIOError: [Errno 11] Resource temporarily unavailable
``` 

And over SSL:

```python
...
>>> client.connect(url, port, keepalive=0)
... (stack trace)
ssl.SSLWantReadError: The operation did not complete (read) (_ssl.c:1000)
```

Both TLS and WebSocket rely on initial communication beyond the socket connection to establish a connection which makes a timeout of 0 impossible to respect for either protocol.

Using a seperate value which is non-zero seems to make more sense, although the meaning of `keepalive` fits the `settimeout` functionality better, so perhaps only calling it for values greater than 0?

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -4612,7 +4612,7 @@ class Client:
             sock = self._ssl_wrap_socket(sock)
 
         if self._transport == "websockets":
-            sock.settimeout(self._keepalive)
+            sock.settimeout(self._connect_timeout)
             return _WebsocketWrapper(
                 socket=sock,
                 host=self._host,
@@ -4667,7 +4667,7 @@ class Client:
             if getattr(self._ssl_context, 'check_hostname', False):  # type: ignore
                 verify_host = False
 
-        ssl_sock.settimeout(self._keepalive)
+        ssl_sock.settimeout(self._connect_timeout)
         ssl_sock.do_handshake()
 
         if verify_host:


Sporarum:
What is the status on this PR ?

Merging it would help me a ton !

cjavad:
Hi @Sporarum personally we ended up just using a keepalive value that was consistent with the rest of our system instead of disabling this feature, previously we maintained a runtime patch like this https://gist.github.com/cjavad/b804b4fada7ff0e18cf77d43b7f731dd, which can be adapted to also patch out websockets etc.

But i would imagine @'ing a maintainer could get this moving otherwise.

Sporarum:
In our case, we want a long keepalive, but a short connection timeout, which does not seem possible because of this bug

Issue #813, opened 2024-02-09
PINGREQ still going out even when QoS 1 publishes are happening within keep alive period


tsturzl:
# Prerequisites

*Note: You may remove this section prior to submitting your report.*

A small team of volunteers monitors issues. Please help us to help you by making it simple to understand and, if possible,
replicate your issue. Prior to reporting a bug please:

 - [ ] Test the latest release of the library.
 - [ ] Search existing issues.
 - [ ] Read the relevant documentation. 
 - [ ] Review your server configuration and logs.
 - [ ] Consider testing against a different server (e.g. [mqtt.eclipseprojects.io](https://mqtt.eclipseprojects.io/) or [test.mosquitto.org](https://test.mosquitto.org/))
 - [ ] If possible, test using another tool (e.g. [MQTTX](https://mqttx.app/) / [mosquitto_sub](https://mosquitto.org/man/mosquitto_sub-1.html))
     to confirm the issue is specific to this client.
 - [ ] If you are unsure if you have found a bug, please consider asking on [stackoverflow](https://stackoverflow.com/) for a quicker response.

# Bug Description

I have a client configured for a 4 second keep alive, and I'm publishing QoS 1 messages every 500ms. I still see PINGREQ going out in a packet sniffer. This means that keep alive is making unneeded ping requests. This doesn't track the behavior the documentation seems to suggest, nor does it seem to resemble the behavior of other clients such as the paho.mqtt.java client. Obviously this isn't a major behavior issue, but we noticed this behavioral difference while packet sniffing python and Java clients, and it seems python is the outlier in terms of behavior, especially since the documentation seems to suggest that keep alive should only ping in absence of other control messages.

# Reproduction

I'll just provide a test script:

```
import paho.mqtt.client as mqtt
import time

client = mqtt.Client()

def on_disconnected(_client, _userdata, _rc):
    print("DISCONNECTED")

client.on_disconnect = on_disconnected

# enable logging, let it create it's own logger
client.enable_logger(logger=None)

print("connecting...")
client.connect("localhost", 1883, 5)
print("connected")
print("starting loop...")
client.loop_start()
print("loop started")
while True:
    print("publishing message...")
    client.publish("testtopic1", "hello world", qos=1)
    time.sleep(0.5)
```

# Environment

* Python version: 3.10.9
* Library version: 1.6.1
* Operating system (including version): Ubuntu 22.04.03
[...]

PR #903, opened 2026-01-20
PINGREQ still going out even when publishes are happening within keep alive period

neel-infocusp:
- Added fix to not send PINGREQ even when publishes are happening (Reference issue: #813 )

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -3238,6 +3238,10 @@ class Client:
                 else:
                     # We haven't finished with this packet
                     self._out_packet.appendleft(packet)
+
+                # Update the last message out time once packet (partial packet) is sent
+                with self._msgtime_mutex:
+                    self._last_msg_out = time_func()
             else:
                 break
 
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -3180,6 +3180,10 @@ class Client:
                 return MQTTErrorCode.MQTT_ERR_CONN_LOST
 
             if write_length > 0:
+                # Update the last message out time once packet (partial packet) is sent
+                with self._msgtime_mutex:
+                    self._last_msg_out = time_func()
+
                 packet['to_process'] -= write_length
                 packet['pos'] += write_length
 
@@ -3238,10 +3242,6 @@ class Client:
                 else:
                     # We haven't finished with this packet
                     self._out_packet.appendleft(packet)
-
-                # Update the last message out time once packet (partial packet) is sent
-                with self._msgtime_mutex:
-                    self._last_msg_out = time_func()
             else:
                 break
 


JamesParrott:
@PierreF This PR should fail review and be rejected.  The original issue (#813) is not an unreasonable feature request, (even though [the spec](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901045) clearly states:

> The Client can send PINGREQ at any time, irrespective of the Keep Alive value, and check for a corresponding PINGRESP to determine that the network and the Server are available. 

and this PR looks simple on the surface.  But there are no tests for it, and it should be closed.

The original issue reasonably points out, that keep alive PINGREQs are unnecessary, when PUBACK or PUBREC responses are being recieved.  But this is only true for QoS 1 and 2 messages.  This PR affects QoS 0 messages too, and introduces a breaking change for them.  

Under this PR, a client making a large write, will delay the keep alive PINGREQ until after it finishes writing all the data to the socket even if nothing's listening on the end of it.  This would seem astonishing to me as a user, especially if I've deliberately set that keep alive timeout to a shorter time period than the anticipated upload time.

PierreF:
I don't understand the fix from the issue description. I'm not sure to understand which bug we try to fix here. With or without this PR, I still have issue #813 using the reproduce method.

The issue #813 is about send pingreq even when there is a flow of PUBLISH / PUBACK. This change only impact the outgoing path (PUBLISH) and slightly change the meaning of _last_msg_out from last message (== MQTT packet) out, to last bytes out.

Could you explain what issue this change fix ?

PierreF:
I had an issue in my test process (I tested the same version, not the PR). I do have a change with your PR and it fix #813.
It do change a bit the (expected) meaning of _last_msg_out from the expected last MQTT packet sent to last bytes out. But currently on main it's just broken and had no meaning (it's never updated on successful packet write).

@JamesParrott I'm not sure what is broken with QoS 0 ?

I partially agree with "client making a large write, will delay the keep alive PINGREQ". Pingreq will be sent if not packet are received (regardless of _last_msg_out). On [condition](https://github.com/eclipse-paho/paho.mqtt.python/blob/7a3d1618bc446a4c63c35ea165b099f6bd95c26b/src/paho/mqtt/client.py#L3276) that trigger a pingreq, only the sending part will delay pingreq due to this change, the receiving part of the condition will still trigger the pingreq. Which is totally wanted, as you said: we do want to emit a pingreq to be sure someone is listenning on the other end.

I think we should make a better change:
* We should fix the meaning we expect for `_last_msg_out`. I think meaning an MQTT packet is full sent is the expected meaning
* Therefore we should only update it as soon as [to_process == 0](https://github.com/eclipse-paho/paho.mqtt.python/blob/7a3d1618bc446a4c63c35ea165b099f6bd95c26b/src/paho/mqtt/client.py#L3191), which will cover the disconnect case
* But I don't get why we update it when [write failed](https://github.com/eclipse-paho/paho.mqtt.python/blob/7a3d1618bc446a4c63c35ea165b099f6bd95c26b/src/paho/mqtt/client.py#L3191), and strongly think it should be removed.

JamesParrott:
> I'm not sure what is broken with QoS 0 ?

My concern is this PR will cause QoS 0 Clients to self disconnect.  Clients publishing under QoS 0 don't receive PUBACKs, so pinging shouldn't be "turned off" for them.  

I now realise that there is other code by which such a QoS 0 Client will still ping the broker (based on self._last_msg_in being too long ago).  But if the broker is delayed, or if the download speed is slow, or for whatever reason if loop_misc gets called again, before the PINGRESP is received, it alls _check_keepalive again, self._last_msg_in has exactly the same value as before (and now it was even longer ago), then as self._ping_t is now !=0 (from the PINGREQ), the code path will hit [`self._sock_close()`](https://github.com/eclipse-paho/paho.mqtt.python/blob/7a3d1618bc446a4c63c35ea165b099f6bd95c26b/sClrc/paho/mqtt/client.py#L3291).

Maybe I'm wrong (that's certainly possible)  or maybe the above just can't happen.  But this is a potential breaking change with a large impact, and at the very least there should be a test for this QoS 0 scenario (plus in my opinion, stronger justification than the "nice to have" of avoiding unnecessary pings any compliant MQTT Client  is are quite entitled to make).

JamesParrott:
OK, I tested this PR a little bit and couldn't create a self disconnect.  

What I should've realised before is that when payload uploads are complete, QoS 1 and 2 publishers should receive PUBREC, PUBREL or PUBACK from the broker, which resets `self._last_msg_in`.  QoS 0 publishers that don't receive those packets, have only existing mechanisms to update `self._last_msg_in`, so eventually should still ping.

It's your call about the risk/value @PierreF .  I've thrown a lot at this, and I have to admit this PR's withstood it all.  It's unreasonable of me to stand in the way now - I withdraw all my objections.

PR #909, opened 2026-04-02
Fix old on_subscribe example in callback migration documentation

sid3windr:
I found a copy/paste mistake in the example of the old code; I am fairly sure what I put instead is correct but feel free to correct me if wrong...

### DIFF
--- a/docs/migrations.rst
+++ b/docs/migrations.rst
@@ -132,7 +132,7 @@ on_subscribe
                 # error processing
 
     # OLD code for MQTTv5
-    def on_disconnect(client, userdata, mid, reason_codes, properties):
+    def on_subscribe(client, userdata, mid, granted_qos, rc, properties=None):
         for sub_result in reason_codes:
             if sub_result == 1:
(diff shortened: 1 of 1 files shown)


docs/migrations.rst
```diff
@@ -132,7 +132,7 @@ on_subscribe
                 # error processing
 
     # OLD code for MQTTv5
-    def on_disconnect(client, userdata, mid, reason_codes, properties):
+    def on_subscribe(client, userdata, mid, granted_qos, rc, properties=None):
```
PierreF, line 135: I believe in MQTTv5, the `granted_qos` was absent (see calling code in [v1.6.1](https://github.com/eclipse-paho/paho.mqtt.python/blob/v1.6.1/src/paho/mqtt/client.py#L3259)).

That's probably why I made a copy/paste mistake, because in MQTTv5 there is no migration for on_subscribe. But I think it still better to explicitly show old code for MQTTv5, even if unchanged for migration.

PR #918, opened 2026-06-11
Pass mypy checks (in lint job, via tox)

### LINKED ISSUE #917: The Mypy check is failing (during the lint job, run via tox)
Most recent run on master:
https://github.com/eclipse-paho/paho.mqtt.python/actions/runs/27304311197

```
lint: commands[3]> mypy src
src/paho/mqtt/client.py:1698: error: Item "None" of "socket | None" has no attribute "recv"  [union-attr]
src/paho/mqtt/client.py:1781: error: Incompatible types in assignment (expression has type "bytes | bytearray", variable has type "bytes")  [assignment]
src/paho/mqtt/client.py:2233: error: Incompatible types in assignment (expression has type "bytes | bytearray", variable has type "bytes")  [assignment]
src/paho/mqtt/client.py:3444: error: Argument 2 to "_packet_queue" of "Client" has incompatible type "bytearray"; expected "bytes"  [arg-type]
src/paho/mqtt/client.py:3572: error: Argument 2 to "_packet_queue" of "Client" has incompatible type "bytearray"; expected "bytes"  [arg-type]
src/paho/mqtt/client.py:3610: error: Argument 2 to "_packet_queue" of "Client" has incompatible type "bytearray"; expected "bytes"  [arg-type]
src/paho/mqtt/client.py:3652: error: Argument 2 to "_packet_queue" of "Client" has incompatible type "bytearray"; expected "bytes"  [arg-type]
src/paho/mqtt/client.py:3701: error: Argument 2 to "_packet_queue" of "Client" has incompatible type "bytearray"; expected "bytes"  [arg-type]
src/paho/mqtt/client.py:4921: error: Incompatible types in assignment (expression has type "bytearray", variable has type "bytes")  [assignment]
Found 9 errors in 1 file (checked 11 source files)
lint: exit 1 (4.45 seconds) /home/runner/work/paho.mqtt.python/paho.mqtt.python> mypy src pid=3097
```

Image

JamesParrott:
Fixes #917, #919, & #920

<edit>  Created PRs to fix those two test failures, and rebased this PR onto them both.  ~~I'm not sure why there is a test failure on tear down (on Python 3.8 only) - I'll look into it.  The test 'failure' on Python 3.7 is probably because Github's setup Python action no longer supports Python 3.7~~

There is only one non-trivial change to run time code - coercing `result` to bytes from bytesarray, to ensure `_WebSocketWrapper._recv_impl`'s returned values really are bytes (its declared return type).  The two trivial run time changes are to re order a simple `and` expression to assist mypy, and to import `overload` from typing.  All other changes are to type annotations only.

Before I renamed the branch:
image

 - SocketLike.send now explicitly accepts bytearray
 - An expression was reordered so that mypy could infer self._socket was not None
 - _force_bytes was generalised, and given some overloads, to assist mypy's type narrowing
 - MQTTMessage.payload and Client._will_payload have been annotated to explicitly also support bytesarray now, as well as bytes.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       max-parallel: 4
       matrix:
-        python: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
     runs-on: ubuntu-latest
     timeout-minutes: 5
     steps:
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -3,7 +3,7 @@ on:
   pull_request:
     branches: [master]
   push:
-    branches: [master, "Fix_test_failure_on_Python_3.7", "Fix_test_failure_on_Python_3.8"]
+    branches: [master]
 jobs:
   tox:
     strategy:
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -3,7 +3,7 @@ on:
   pull_request:
     branches: [master]
   push:
-    branches: [master]
+    branches: [master, Fix_test_failure_on_Python_3.8]
 jobs:
   tox:
     strategy:
@@ -20,13 +20,20 @@ jobs:
           repository: eclipse/paho.mqtt.testing
           ref: a4dc694010217b291ee78ee13a6d1db812f9babd
           path: paho.mqtt.testing
+          
+      - uses: LizardByte/actions/actions/setup_python@v2026.605.34721
+        if: ${{ matrix.python == '3.7' }}
+        with:
+          python-version: ${{ matrix.python }}
       - uses: actions/setup-python@v5
+        if: ${{ matrix.python != '3.7' }}
         with:
           python-version: ${{ matrix.python }}
           cache: pip
           cache-dependency-path: |
             tox.ini
             setup.py
+
       - run: pip install tox
       - run: tox -e py
       - uses: codecov/codecov-action@v3
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -3,7 +3,7 @@ on:
   pull_request:
     branches: [master]
   push:
-    branches: [master, Fix_test_failure_on_Python_3.8]
+    branches: [master, "Fix_test_failure_on_Python_3.7", "Fix_test_failure_on_Python_3.8"]
 jobs:
   tox:
     strategy:
@@ -20,7 +20,7 @@ jobs:
           repository: eclipse/paho.mqtt.testing
           ref: a4dc694010217b291ee78ee13a6d1db812f9babd
           path: paho.mqtt.testing
-          
+
       - uses: LizardByte/actions/actions/setup_python@v2026.605.34721
         if: ${{ matrix.python == '3.7' }}
         with:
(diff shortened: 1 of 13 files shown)


.github/workflows/tox.yml
```diff
@@ -20,13 +20,20 @@ jobs:
           repository: eclipse/paho.mqtt.testing
           ref: a4dc694010217b291ee78ee13a6d1db812f9babd
           path: paho.mqtt.testing
+
+      - uses: LizardByte/actions/actions/setup_python@v2026.605.34721
```
PierreF: If I understand the aim of this change (supporting Python 3.7) I don't think we should.
It's EOL and too old, I'm in the way of removing it (#916) but I missed the Python version list in this file. I'll fix the list of Python in my PR

tests/lib/conftest.py
```diff
@@ -71,7 +71,9 @@ def starter(name: str, expected_returncode: int = 0) -> None:
 
         def fin():
             stop_process(proc)
-            if proc.returncode != expected_returncode:
+            # Return code can be None on Python 3.8
+            # https://docs.python.org/3.8/library/subprocess.html#subprocess.Popen.returncode
```
PierreF: It's not Python 3.8 specific. Any Python version (at least it's still documented on Python 3.14) could return None

```suggestion
            # https://docs.python.org/library/subprocess.html#subprocess.Popen.returncode
```

PierreF (changes requested):
Thanks for that pass, it would have been my next tasks after #916.

tests/lib/conftest.py
JamesParrott: Oh you're right.  I should've checked - sorry.  There must be some other reason(s) this error only occurred on 3.8

.github/workflows/tox.yml
JamesParrott: Sure.  It's a lot simpler to drop support for Python 3.7 (and 3.8 too) if you'd prefer.  I'd support that personally.  But it's a bigger, breaking change.  I didn't presume to make that call.

N.B.  I didn't immediately realise this.  "Drop" sounds bad, but people stuck on older Pythons, will just be stuck on older versions of Paho.mqtt.

src/paho/mqtt/client.py
```diff
@@ -595,7 +600,7 @@ def __init__(self, mid: int = 0, topic: bytes = b""):
         self.mid = mid
         """ The message id (int)."""
         self._topic = topic
-        self.payload = b""
+        self.payload: bytes|bytearray = b""
```
pswsm: Not really important, but to keep the style:
```suggestion
        self.payload: bytes | bytearray = b""
```

src/paho/mqtt/client.py
```diff
@@ -3758,7 +3763,7 @@ def _messages_reconnect_reset(self) -> None:
     def _packet_queue(
         self,
         command: int,
-        packet: bytes,
+        packet: bytes|bytearray,
```
pswsm: Same thing in here, missing spaces between the union:
```suggestion
        packet: bytes | bytearray,
```

pswsm (commented):
Minor style diffs, feel free to ignore them.

src/paho/mqtt/client.py
JamesParrott: @pswsm Thanks!  I just copied a prior example from elsewhere in the file.  

Other than these, I think there's plenty enough in this PR already.  

Moving forwards, all styling issues will be fixed by running Black on the whole code base.  This will also fix the Black job in the lint workflow.

.github/workflows/tox.yml
JamesParrott: LizardByte conditional step removed

pswsm (approved):

PR #922, opened 2026-06-11
Fix test failure on python 3.7

### LINKED ISSUE #919: Spurious test failures on Python 3.7
This is the typical 'failure' seen all the time when testing on EOL Python versions.  I'm guessing Github's setup-python action, and/or Ubuntu 24.04 have dropped Python 3.7 support.

Image

Some possible fixes that worked for me in the past are:
- Run the action in a Python 3.7 docker container, instead of directly on the Ubuntu host
 - LizardByte/actions/actions/setup-python instead (uses pyenv under the hood)
 - For newer versions thatn 3.7, installing the required Python with uv works, but I don't believe they support 3.7 any more.
 - Perhaps use tox to create its required Python envs itself?

Or drop support for Python 3.7 of course.

JamesParrott:
Fixes https://github.com/eclipse-paho/paho.mqtt.python/issues/919

Uses LizardByte setup Python action to setup Python 3.7 (with pyenv)

image

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -3,7 +3,7 @@ on:
   pull_request:
     branches: [master]
   push:
-    branches: [master]
+    branches: [master, Fix_test_failure_on_Python_3.8]
 jobs:
   tox:
     strategy:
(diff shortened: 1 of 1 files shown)


JamesParrott:
Included in PR #918

Issue #914, opened 2026-05-28
unexpected MQTT_ERR_KEEPALIVE when chunk cannot be received within keepalive interval


Uezzi:
When trying to request one large message (~100 MB) from mqtt broker, connection is reset with MQTT_ERR_KEEPALIVE error when chunk cannot be received within double of configured keepalive interval.

Using following pseudo code:
```
import paho.mqtt.client as mqtt
client = mqtt.Client()
client.connect(my_url, keepalive=30)
```

When chunk is not received within 60 seconds, MQTT_ERR_KEEPALIVE will occur.

Reason: Keep-Alive timeout detection is done by [_check_keepalive()](https://github.com/eclipse-paho/paho.mqtt.python/blob/v2.1.0/src/paho/mqtt/client.py#L3261) function using `self._last_msg_out` and `self._last_msg_in`.
`self._last_msg_in` is properly updated after a [full packed has been received](https://github.com/eclipse-paho/paho.mqtt.python/blob/v2.1.0/src/paho/mqtt/client.py#L3157) and after [reading 100 parts](https://github.com/eclipse-paho/paho.mqtt.python/blob/v2.1.0/src/paho/mqtt/client.py#L3137) of a huge packet. However there might happen a `BlockingIOError` when calling `self._sock_recv()`. This happens with my setup regularly. In this case `self._last_msg_in` is not updated, resulting in keepalive error.

Suggested fix: update `self._last_msg_in` regularly whenever [data is received](https://github.com/eclipse-paho/paho.mqtt.python/blob/v2.1.0/src/paho/mqtt/client.py#L3133) from socket.

JamesParrott:
A MQTT_ERR_KEEPALIVE error is to be expected within 30 - 60 seconds when creating a connection with `client.connect(my_url, keepalive=30)`.  If you need longer, pass a bigger value to `keepalive`.  

@PierreF I'd close this one.  It's user error at best, AI at worst.

Uezzi:
> A MQTT_ERR_KEEPALIVE error is to be expected within 30 - 60 seconds when creating a connection with `client.connect(my_url, keepalive=30)`. If you need longer, pass a bigger value to `keepalive`.

Why should I use a bigger `keepalive` value when I receive data all the time (in the 30..60 seconds you mentioned)?
The keepalive detection doesn't work because of the way `BlockingIOError` is handled at the moment (without updating `self._last_msg_in`).

JamesParrott:
[...]

PR #923, opened 2026-06-15
Send multiple pings during large downloads, resetting self._ping_t

JamesParrott:
Hi everyone, I'm very interested, but am not too invested in this  PR - I'm keen to hear everyone's thoughts.  All feedback is welcome.  First and foremost, I think the intention of the spec to fix this problem (MQTT being a Lightweight messaging protocol) is that clients should pick a large enough keep alive interval, to cover the size of downloads they expect to receive on topics they're subscribed to.  However subscribers don't control what others publish, and a broker may require a shorter keepalive than the client requested.

From local testing, on a mini slow network connection (between my dev laptop and a Beaglebone Black over a USB cable in my case), I can't speak for @Uezzi but I've found this PR fixes issue #914.

There are two separate checks that must be passed in `loop_misc` to avoid the client disconnecting itself (calling self._sock_close).  
 - [1)](https://github.com/eclipse-paho/paho.mqtt.python/blob/c4af656b902baf36f721a06634c6f1d6af182ddc/src/paho/mqtt/client.py#L3291), if we are connected, but pingreq(s) have gone out and a ping response has not been received after the most recent one left (self._ping_t == 0), and either no other msg packets have been sent within the keep alive, or none have been received during the keep alive interval.  If a pingreq has not gone out recently, this check also sends one.
 - [2)](https://github.com/eclipse-paho/paho.mqtt.python/blob/c4af656b902baf36f721a06634c6f1d6af182ddc/src/paho/mqtt/client.py#L2156) if a Ping Response has not been received within the keep alive interval of the last Ping request going out.

It looks to me like this second one is a straightforward implementation of "If a Client does not receive a PINGRESP packet within a reasonable amount of time after it has sent a PINGREQ, it SHOULD close the Network Connection to the Server." from the spec.

Within the first check however (you may have already spotted it) is an edge case that forms a grey area - what should the client do if a ping req has gone out, no other packets have gone out since during the keep alive interval, but in that time it's been happily downloading the partial msg packets that form another packet.  

The server / broker may well have sent a PINGRESP, but currently on a slow network, for large downloads and short keep alive intervals, it  is possible that the client does not read it in time, to avoid either self disconnection mechanism above.  The connection may be fine, but the PINGRESP may simply be stuck behind too many unread partial packets of a large payload.  About 100MB to 250MB can recreate precisely this situation with a 30 second keep alive.  

In this situation, before the large, slow download completes, a Paho.mqtt.Python client will disconnect itself.  Regardless of the keepalive values they set and/or received, some users many find this astonishing.  The spec (v5) does state:  

> If a Client does not receive a PINGRESP packet within a reasonable amount of time after it has sent a PINGREQ, it SHOULD close the Network Connection to the Server.

The essence of this PR is to change the definition of "reasonable amount of time" from the keep alive interval, to "the maximum of the keep alive interval and however much time is needed to complete the current download"  The maximum payload size is ~ 255MB, and as long as the network connection can pass traffic at an average speed v > 0, there is a finite upper bound of ~255 / v.  for this time.  Furthermore, it is likely not an unreasonable amount of time, in relation to the expectations and experience of users of slow networks.

This PR fixes this simply by altering current behaviour, to ping again, even if a prior ping request has not received a ping response.  This also resets `self._ping_t` avoiding the second check.
The two new situations when (if this PR is accepted) we would now ping again, even when no ping response has been received since the most recent ping req:
 - i) If the last message out was too long ago (beyond keep alive), but the Client knows other packets have been received.  In this case we ping again to avoid the server kicking the client for inactivity, resetting self._ping_t.  The check whether the Client should disconnect due to no ping resp  being received for the previous pingreq is left to 2) with the latest value of self._ping_t, once no packets have been received for longer than self._keepalive.
  - ii) if the last packet in was too long ago (beyond keep alive).  A ping resp is a normal packet, so has not been received recently enough either.  But this case will also be (or would have already been) handled by 2).


pswsm:
After reading the MQTT specification, the part relating to the Keep Alive, this is still in-spec IMO. As you've stated, the spec says "SHOULD", and this is a valid reason to skip this recommendation, I think. I don't feel like I'm the most qualified guy to make the final decision, though.

.github/workflows/build.yml
```diff
@@ -1,9 +1,9 @@
 name: build
 on:
   pull_request:
-    branches: [master]
+    branches: [master, Do_not_self_disconnect_during_large_download]
```
PierreF:
```suggestion
    branches: [master]
```
(apply to other similar changes)

I'm not sure what the intend.
1) it's shouldn't work (i.e. shouldn't change CI behavior on this PR), or else there is a security issue with the CI
2) it's shouldn't be necessary to run CI jobs on PR. AFAIK they run because this is a PR *targeting* the branch master (and it's not your very first contribution)
3) If this change is really needed, we can't hard-code the same of a branch

PierreF (changes requested):
I've re-read the MQTT spec and agree with your conclusion:
* The client *must* send Ping request every keep-alive interval if not sending MQTT packet (we already did this, and continue to do this)
* The client should disconnect after *reasonable* delay after sending pingreq if pingresp isn't received.
  * If data is actively receive, I do believe the reasonable delay could be multiple minutes


Maybe a infinite delay is a bit too much (well depend on how slow the download is, but at 1400 bytes every 10 seconds, 255 MiB is 22 days ;)), but I think have a secondary delay (ping non-responded pingresp) and secondary settings (max keep-alive delay during large download) feels like too complex but this corner case.

PierreF:
I think following test should pass but don't... it might a bit a "test" issue, but unsure.

Your PR rely on `_last_msg_in` being updated "soon" after any bytes received on the socket, however during large download, it only get updated once [100 retry](https://github.com/eclipse-paho/paho.mqtt.python/blob/7a3d1618bc446a4c63c35ea165b099f6bd95c26b/src/paho/mqtt/client.py#L3142) were done, and its much more likely to reach the [BlockingIO](https://github.com/eclipse-paho/paho.mqtt.python/blob/7a3d1618bc446a4c63c35ea165b099f6bd95c26b/src/paho/mqtt/client.py#L3129) error (only way to not reach it, is the network send data fast enough to always fill the socket buffer, but slow enough to hit the download-too-large issue).

I do think we should review a bit when `_last_msg_in` is updated... and maybe renamed it (or add new tracker) `_last_bytes_received`. It would be more clear that is means "any bytes received from network". `_last_msg_in` sound more like "last MQTT packet received" - which is not what we need for this PR.

The test I've try to write (I hope it's correct, but can't guarantee):

```
class TestLargeDownload:
    def test_no_disconnect_during_large_download(self, fake_broker: FakeBroker) -> None:
        mqttc = client.Client(
            CallbackAPIVersion.VERSION2,
            "test_no_disconnect_during_large_download",
            transport=fake_broker.transport,
        )

        payload_data = b"very Very Big big payload"
        is_connected = threading.Event()
        is_subscribed = threading.Event()
        publish_received = threading.Event()

        def on_connect(mqttc, obj, flags, rc, properties):
            assert rc == 0
            is_connected.set()

        def on_subscribe(cl, userdata, mid, reason_code_list, properties):
            is_subscribed.set()
        
        def on_message(cl, userdata, message):
            assert isinstance(cl, client.Client)
            assert isinstance(message, client.MQTTMessage)
            assert message.payload == payload_data
            publish_received.set()

        mqttc.on_connect = on_connect
        mqttc.on_subscribe = on_subscribe
        mqttc.on_message = on_message

        mqttc.enable_logger()
        mqttc.connect("localhost", fake_broker.port, keepalive=1)
        mqttc.loop_start()

        try:
            fake_broker.start()

            connect_packet = paho_test.gen_connect(
                "test_no_disconnect_during_large_download", keepalive=1)
            fake_broker.expect_packet("connect", connect_packet)

            connack_packet = paho_test.gen_connack(rc=0)
            count = fake_broker.send_packet(connack_packet)
            assert count == len(connack_packet)

            is_connected.wait(timeout=5)
            rc, mid = mqttc.subscribe("topic")
            assert rc == MQTTErrorCode.MQTT_ERR_SUCCESS
            assert mid is not None

            subscribe_packet = paho_test.gen_subscribe(mid, "topic", 0)
            fake_broker.expect_packet("subscribe", subscribe_packet)

            suback_packet = paho_test.gen_suback(2, 0)
            count = fake_broker.send_packet(suback_packet)
            assert count == len(suback_packet)

            is_subscribed.wait(timeout=5)

            publish_packet = paho_test.gen_publish("topic", 0, payload_data, mid=3)
            assert len(publish_packet) >= 32
            publish_packet_1 = publish_packet[:5] # Enough to cover the packet type
            publish_packet_2 = publish_packet[5:10]
            publish_packet_3 = publish_packet[10:15]
            publish_packet_4 = publish_packet[15:20]
            publish_packet_5 = publish_packet[20:25]
            publish_packet_6 = publish_packet[25:]

            count = fake_broker.send_packet(publish_packet_1)
            assert count == len(publish_packet_1)
            time.sleep(0.8)

            count = fake_broker.send_packet(publish_packet_2)
            assert count == len(publish_packet_2)
            time.sleep(0.8)

            # 1.6 seconds elapsed, client MUST have sent pingreq
            ping_req_packet = paho_test.gen_pingreq()
            fake_broker.expect_packet("pingreq", ping_req_packet)

            count = fake_broker.send_packet(publish_packet_3)
            assert count == len(publish_packet_3)
            time.sleep(0.8)

            count = fake_broker.send_packet(publish_packet_4)
            assert count == len(publish_packet_4)
            time.sleep(0.8)

            # Another 1.6 seconds elapsed, client MUST have sent another pingreq
            ping_req_packet = paho_test.gen_pingreq()
            fake_broker.expect_packet("2nd pingreq", ping_req_packet)

            count = fake_broker.send_packet(publish_packet_5)
            assert count == len(publish_packet_5)
            time.sleep(0.8)

            count = fake_broker.send_packet(publish_packet_6)
            assert count == len(publish_packet_6)

            publish_received.wait(timeout=5)
        finally:
            mqttc.loop_stop()
            fake_broker.finish()
```

.github/workflows/build.yml
JamesParrott: Apologies.  I should've reverted that branch name.  Is it possible to relax the branches: [master] everywhere, so we can work on other branches on our forks?

Uezzi:
@PierreF 
> Your PR rely on `_last_msg_in` being updated "soon" after any bytes received on the socket, however during large download, it only get updated once [100 retry](https://github.com/eclipse-paho/paho.mqtt.python/blob/7a3d1618bc446a4c63c35ea165b099f6bd95c26b/src/paho/mqtt/client.py#L3142) were done, and its much more likely to reach the [BlockingIO](https://github.com/eclipse-paho/paho.mqtt.python/blob/7a3d1618bc446a4c63c35ea165b099f6bd95c26b/src/paho/mqtt/client.py#L3129) error (only way to not reach it, is the network send data fast enough to always fill the socket buffer, but slow enough to hit the download-too-large issue).

I observed exactly this issue with BlockingIOError using my setup - see #914 .

> I do think we should review a bit when `_last_msg_in` is updated... and maybe renamed it (or add new tracker) `_last_bytes_received`. It would be more clear that is means "any bytes received from network". `_last_msg_in` sound more like "last MQTT packet received" - which is not what we need for this PR.

Renaming to `_last_bytes_received` sounds good!

src/paho/mqtt/client.py
```diff
@@ -3274,7 +3276,7 @@ def _check_keepalive(self) -> None:
             last_msg_in = self._last_msg_in
 
         if self._sock is not None and (now - last_msg_out >= self._keepalive or now - last_msg_in >= self._keepalive):
-            if self._state == _ConnectionState.MQTT_CS_CONNECTED and self._ping_t == 0:
+            if self._state == _ConnectionState.MQTT_CS_CONNECTED and (self._ping_t == 0 or now - last_msg_in < self.keepalive):
```
PierreF: As said in https://github.com/eclipse-paho/paho.mqtt.python/issues/914#issuecomment-4793844846, I think `last_msg_in` should be renamed and reworked into `last_bytes_received`. It should count the last time a bytes was received on the network (i.e. as soon as `self._sock.recv` return a length >= 1).

But when the change of `last_bytes_received` is done, I think it will meet required the translation of MQTT spec I write in #914:
* We must still send PINGREQ (or any MQTT packet) every keepalive interval: the `or` added by this PR will cause a new PINGREQ to be sent (and since PINGREQ update lsg_msg_out, we won't flood PINGREQ)
* We should not disconnect if bytes arrive on the network: this is the primary purpose of the `or` added
* We should still disconnect if nothing arrive from the network: this `or` is capped by keepalive

src/paho/mqtt/client.py
```diff
@@ -3274,7 +3276,7 @@ def _check_keepalive(self) -> None:
...
                 try:
                     self._send_pingreq()
```
PierreF: while writting above comment I realized one point which is unclear if it really hold:
> since PINGREQ update lsg_msg_out, we won't flood PINGREQ

I'm not totally sure this is true. Because it only true if loop_write is called, which if behind a condition (`if self._thread is None and self._on_socket_register_write is None`). I do believe however that `self._ping_t` is always updated (OR we are disconnected / will disconnect, and therefore won't loop here).

So I think we should protect this call to pingreq:

```suggestion
                    if self._ping_t  == 0 or now - self._ping_t >= self._keepalive:
                        self._send_pingreq()
```

In any case, I don't see a protocol requirement to send a PINGREQ less than keepalive after previous one. The only requirement is that we need to send any MQTT packet every keepalive, to ensure broker don't disconnect us. But we are never required to send PINGREQ sooner.

PierreF (changes requested):

src/paho/mqtt/client.py
JamesParrott: >  I think last_msg_in should be renamed and reworked into last_bytes_received

You're probably right.  That would be a lot better, long term.   Shall I close this PR?

PierreF: No, you could update this PR. The rename & fix (i.e updating it on right place) of last_msg_in -> last_bytes_received is only a prerequisite for your fix on large download. I believe it suited to fix both of them together.

src/paho/mqtt/client.py
JamesParrott: Yes, it's more robust to put this back in, to handle all the possible ways the can effectively choose, to handle calling loop_write.  It essentially rate-limits PINGREQs.

JamesParrott:
I've:
- renamed `Client._last_msg_in` to `Client._last_bytes_read` and 
- added he extra condition check to rate limit PINGREQs
(as you suggested, Pierre).

The current implementation does only update `Client._last_bytes_read` every 100 reads, so I added a comment
```py
             if count == 0:
                # self._last_bytes_received could be updated on every iteration
                # (also when count > 0).  Not doing so avoids any
                # performance overhead from acquiring the lock every iteration.
                with self._msgtime_mutex:
                    self._last_msg_in = time_func()
                    self._last_bytes_received = time_func()
                return MQTTErrorCode.MQTT_ERR_AGAIN
```
Would it interfere with the main thread to acquire the lock once, outside the loop?  

Alternatively if we really do only need to update `._last_bytes_read` after the loop finishes, the update code at the end of that method can be shared with a little refactoring of the logic returning from the loop:  https://github.com/JamesParrott/paho.mqtt.python/blob/5f7d0e542ab41c735d3ba6506276ee4e8521b02d/src/paho/mqtt/client.py#L3149

PierreF:
By not updating it on every loop, I think we have a issue with some reads work (like first 2) than "timeout" (we have a timeout of 0 - setblocking(False), meaning raise BlockingIOError if data aren't immediately available).
Therefore on large download package, I fear most of the time few loop of _packet_read will succeed, but after we will receive a BlockingIOError -> return MQTT_ERR_AGAIN and don't update _last_bytes_received.

Should we really try to optimize to avoid one lock ? On my system is seems that acquiring a lock cost 38 ns, so optimizing this seems irrelevant since the socket.recv took 10x that time (532ns)

I would update _last_bytes_received everytime to avoid the issue and not trying further optimization since they don't seems necessary

```
>>> timeit.timeit('mutex.acquire(); mutex.release()', 'import threading; mutex = threading.Lock()')
0.03802662482485175

>>> init_code="""import socket
srv = socket.socket()
srv.bind(('127.0.0.1', 0))
srv.listen(1)
client = socket.socket()
client.connect(srv.getsockname())
server_socket, peer = srv.accept()
server_socket.setblocking(False)
"""

>>> bench_code="""try:
    server_socket.recv(1)
except BlockingIOError:
    pass
"""

>>> timeit.timeit(bench_code, init_code)
0.532457958906889
```

JamesParrott:
Agreed.  I would have thought the code was IO bound, not CPU bound anyway.

PierreF (approved):

PR #926, opened 2026-07-11
docs: document exceptions raised by MQTTMessageInfo.is_published

### LINKED ISSUE #906: Please document exceptions raised by MQTTMessageInfo.is_published
`MQTTMessageInfo.is_published()` can raise exceptions but this is not properly documented.  
The following should probably be copied from `MQTTMessageInfo.wait_for_publish()`:

```
        :raises ValueError: if the message was not queued due to the outgoing
            queue being full.

        :raises RuntimeError: if the message was not published for another
            reason.
``` 

Also, I think that the current description is ambiguous especially regarding the meaning of a `False` result that can easily be misunderstood as a failure to publish. 

```Returns True if the message associated with this object has been published, else returns False.```
 
I propose something like that:

```Returns False while the message associated with this object is still being processed or True after it has been published.```

Puneet15556:
Fixes #906

`is_published()` raises the same `ValueError`/`RuntimeError` as `wait_for_publish()`, but its docstring did not mention them. This PR copies the `:raises:` documentation over and clarifies that a `False` return means publication has not completed yet — not that it failed.

No behavior change; docstring only.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -574,7 +574,15 @@ class MQTTMessageInfo:
         """Returns True if the message associated with this object has been
         published, else returns False.
 
-        To wait for this to become true, look at `wait_for_publish`.
+        Returning False does not mean the publication failed; it means the
+        publication has not completed yet. To wait for this to become true,
+        look at `wait_for_publish`.
+
+        :raises ValueError: if the message was not queued due to the outgoing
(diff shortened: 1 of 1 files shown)


JamesParrott:
LGTM.  Docstring change only (as the author of the issue requested).

JamesParrott (approved):
All good

Issue #927, opened 2026-07-13
Expired SSL Cert causing 4 tests to fail in CI


JamesParrott:
When `tests/ssl/server.crt` was created (presumably just over 5 years ago), it was given a life  of 5 years, which expired a week ago
https://github.com/eclipse-paho/paho.mqtt.python/blob/a734a9c7ee6b1523a7b6b98e0b9385a040b28fd2/tests/ssl/server.crt#L9

```
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number: 1 (0x1)
        Signature Algorithm: sha256WithRSAEncryption
        Issuer: C=GB, ST=Derbyshire, O=Paho Project, OU=Testing, CN=Signing CA
        Validity
            Not Before: Jul  7 11:14:42 2021 GMT
            Not After : Jul  6 11:14:42 2026 GMT
```

E.g. (fails on all of Python 3.9-3.14) https://github.com/JamesParrott/paho.mqtt.python/actions/runs/29231488075/job/86756485999
```
============================= test session starts ==============================
platform linux -- Python 3.14.6, pytest-9.1.1, pluggy-1.6.0
cachedir: .tox/py/.pytest_cache
rootdir: /home/runner/work/paho.mqtt.python/paho.mqtt.python
configfile: pyproject.toml
testpaths: tests, src
plugins: cov-7.1.0
collected 163 items

collected 163 items

tests/lib/test_01_asyncio.py .                                           [  0%]
tests/lib/test_01_decorators.py .                                        [  1%]
tests/lib/test_01_keepalive_pingreq.py .                                 [  1%]
tests/lib/test_01_no_clean_session.py .                                  [  2%]

...

```

```

tests/test_websocket_integration.py ..................                   [ 97%]
tests/test_websockets.py ....                                            [100%]

==================================== ERRORS ====================================
________________ ERROR at teardown of test_08_ssl_connect_alpn _________________

    def fin():
        stop_process(proc)
        # If return code is None, process had not stopped when a method was last called on proc.
        # https://docs.python.org/library/subprocess.html#subprocess.Popen.returncode
        if proc.returncode != expected_returncode:
>           raise RuntimeError(f"Client {name} exited with code {proc.returncode}, expected {expected_returncode}")
E           RuntimeError: Client 08-ssl-connect-alpn.py exited with code 1, expected 0

tests/lib/conftest.py:85: RuntimeError
----------------------------- Captured stderr call -----------------------------
[...]

PR #931, opened 2026-07-13
Generate ssl certs in a text fixture, delete certs, handle self._sock is None in CLient._loop, run workflows on all branches, sort slots, & preppend underscore to unused variable names.

### LINKED ISSUE #928: Mypy 2.2.0 is failing (Tox lint job) src/paho/mqtt/client.py:1668: error: Value of type variable "_R" of "select" cannot be "SocketLike | None"  [type-var]
```
lint: commands[3]> mypy src
src/paho/mqtt/client.py:1668: error: Value of type variable "_R" of "select" cannot be "SocketLike | None"  [type-var]
src/paho/mqtt/client.py:1668: error: Value of type variable "_W" of "select" cannot be "SocketLike | None"  [type-var]
Found 2 errors in 1 file (checked 11 source files)
lint: exit 1 (5.04 seconds) /home/runner/work/paho.mqtt.python/paho.mqtt.python> mypy src pid=3062
```

E.g. https://github.com/JamesParrott/paho.mqtt.python/actions/runs/29231488140/job/86756486247

mypy 2.2.0:
```
(.venv) C:\...\paho.mqtt.python>mypy --version && mypy src
mypy 2.2.0 (compiled: yes)
src\paho\mqtt\client.py:1668: error: Value of type variable "_R" of "select" cannot be "SocketLike | None"  [type-var]
src\paho\mqtt\client.py:1668: error: Value of type variable "_W" of "select" cannot be "SocketLike | None"  [type-var]
src\paho\mqtt\client.py:4631: error: Module has no attribute "AF_UNIX"; maybe "AF_IPX"?  [attr-defined]
Found 3 errors in 1 file (checked 11 source files)
```

mypy 2.1.0:
```
(.venv) C:\...\paho.mqtt.python>uv pip install mypy==2.1.0 --upgrade
Resolved 6 packages in 576ms
Prepared 1 package in 44ms
Uninstalled 1 package in 905ms
Installed 1 package in 3.77s
 - mypy==2.2.0
 + mypy==2.1.0

(.venv) C:\...\paho.mqtt.python>mypy --version && mypy src
mypy 2.1.0 (compiled: yes)
src\paho\mqtt\client.py:4631: error: Module has no attribute "AF_UNIX"; maybe "AF_IPX"?  [attr-defined]
Found 1 error in 1 file (checked 11 source files)
```

(The "AF_UNIX"; maybe "AF_IPX" error is only because I'm type checking from Windows)

JamesParrott:
Fixes #928 and #927 (hopefully), and gets the CI pipeline working again

 - ~~I generated New test ssl certs using gen.sh (see below for a copy and paste of the shell session I created them from, not that anyone should be relying on the certs).~~
 - Run the build, lint and test (tox) workflows on all branches to make development much easier for contributers
 - Mypy 2.2.0 (unlike 2.2.1) complains (rightly) that self._sock could be None in a call to `select.select`.  Handle this case in Client,_loop by returning MQTT_ERR_NO_CONN.
 - Rename a few unused variables, to make them start with underscores, to make Ruff happy.
 - [Edit] As dotlambda suggested, I added a text fixture (session scoped) to generate SSL certs dynamically using gen.sh in a Pytest tmpdir, and deleted all the expired certs (the tests still pass).


dotlambda (commented):
Why not generate new certificates for each test run, e.g. in `conftest.py`?

dotlambda:
> `c_rehash certs`

That line caused `Skipping certs, can't write` for me because the certs directory doesn't exist (the line creating it is commented out). Can it be removed?

JamesParrott:
> Why not generate new certificates for each test run, e.g. in conftest.py?

That's a great idea for a separate PR.  The pressing need right now is to fix the CI tests, to unblock all the other PRs.

> >  c_rehash certs
> 
> That line caused Skipping certs, can't write for me because the certs directory doesn't exist (the line creating it is commented out). Can it be removed?

I didn't write that line, I just reran the script it's in to regenerate the certs, hopefully in exactly the same way as the previous ones were generated.

If a directory doesn't exist, try running the whole gen.sh - directories are created in it here, at the start:

```bash
for a in root signing; do
       rm -rf ${a}CA/
       mkdir -p ${a}CA/newcerts
       touch ${a}CA/index.txt
       echo 01 > ${a}CA/serial
       echo 01 > ${a}CA/crlnumber
done
```

dotlambda:
> try running the whole gen.sh

I did.



> directories are created in it here, at the start

Not the `certs` directory afaict.

JamesParrott:
Oh sorry I get it now.  Interestingly, [a line to create a certs dir](https://github.com/eclipse-paho/paho.mqtt.python/blob/a734a9c7ee6b1523a7b6b98e0b9385a040b28fd2/tests/ssl/gen.sh#L77) was commented out.

Maybe the `c_rehash` line should've been commented out too?  I can't say I know what it does anyway

dotlambda:
> Interestingly, [a line to create a certs dir](https://github.com/eclipse-paho/paho.mqtt.python/blob/a734a9c7ee6b1523a7b6b98e0b9385a040b28fd2/tests/ssl/gen.sh#L77) was commented out.

Exactly what I said:

> the line creating it is commented out

> Maybe the `c_rehash` line should've been commented out too?

That was how I fixed it on my end.

JamesParrott:
I guess I read that, but I didn't clock it!

> Why not generate new certificates for each test run, e.g. in `conftest.py`?

My focus is on keeping the PR as simple as possible for Pierre.  But on second thoughts, generating certs in a test fixture, would allow us to just get rid of all these certs from source control, which would be nice.

JamesParrott:
@dotlambda I reckon I've got it working.  Would you like co-authorship of the commit, or would you prefer to be left out of it (e.g. remain blameless!)?  https://github.com/JamesParrott/paho.mqtt.python/actions/runs/29363733396

dotlambda:
> Would you like co-authorship of the commit, or would you prefer to be left out of it

Either way is fine. I didn't really contribute besides suggesting the removal of one line ;-)

.github/workflows/lint-python.yml
```diff
@@ -1,9 +1,7 @@
 name: lint_python
 on:
   pull_request:
-    branches: [master]
```
PierreF, line 4: Just to be sure of the objective of this changes:
* This aim to have Github CI running on the contributor fork, for example in https://github.com/JamesParrott/paho.mqtt.python/actions (contributor's space, not eclipse-paho space)
* Said otherwise, it's having CI running on contributor's branch, not just when contributor create the PR ?

If above if true, change looks fine. Else, I believe change isn't needed as never really read the Github workflows documentation, but I always read current config as:
* Run CI on any PR that target the branch "master" (a.k.a. all PR created)
* And run when commit are pushed to branch "master" (a.k.a. when a PR is merged)

src/paho/mqtt/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "2.1.1.dev0"
+__version__ = "2.1.1.dev1"
```
PierreF: Not sure why the bump here. But does matter if we bump or not (so we could leave it)

The ".dev" is only here because it allow Git branch to be more recent that latest release `2.1.0` (latest release) is older than `2.1.1.devN`. And at the same time being oldest that next release (`2.1.1.devN` is older than `2.1.1`).

But therefore excepted that `.devN` is added right after a release, I never change this number.

tests/ssl/gen.sh
```diff
@@ -18,65 +18,66 @@ rm -rf certs
...
 openssl genrsa -out server.key 2048
-openssl req -new -key server.key -out server.csr -config openssl.cnf -subj "${SBASESUBJ}/CN=localhost/"
-openssl ca -batch -config openssl.cnf -name CA_signing -out server.crt -infiles server.csr
+openssl req -new -key server.key -out server.csr -config $CONF_FILE -subj "${SBASESUBJ}/CN=localhost/"
+openssl ca -batch -config $CONF_FILE -name CA_signing -out server.crt -infiles server.csr
 
 # Expired server certificate, based on the above server key.
-openssl req -new -days 1 -key server.key -out server-expired.csr -config openssl.cnf -subj "${SBASESUBJ}/CN=localhost/"
+openssl req -new -days 1 -key server.key -out server-expired.csr -config $CONF_FILE -subj "${SBASESUBJ}/CN=localhost/"
 echo -n > signingCA/index.txt
```
PierreF:
```suggestion
> signingCA/index.txt
```

The `echo -n` isn't working for me, since is result in index containing `-n`. Nothing really new, just that now `gen.sh` is "required" for test.
Having no command to redirect should result file being empty and I believe it's the most portable way.

I think this is a bash / sh difference since echo is a shell built-in and on CI Ubuntu is used (so sh should be ash, which AFAIK is an POSIX sh) when on my system (MacOS) sh is bash.

PierreF (changes requested):
Minor feedback on gen.sh and just a question on workflows.

Other than that LGTM

.github/workflows/lint-python.yml
JamesParrott: You're right - this one can stay targetted to master.  As long as the `on: push:` one is relaxed, that will allow contributors to work on their own branches.

src/paho/mqtt/__init__.py
JamesParrott: Agreed.  It's not useful to bump the `N` in `.devN`.

tests/ssl/gen.sh
JamesParrott: How about `touch some_file.txt` instead of `echo -n > some_file.txt` ?

> SYNOPSIS         [top](https://www.man7.org/linux/man-pages/man1/echo.1.html#top_of_page)
> 
>        echo [SHORT-OPTION]... [STRING]...
>        echo LONG-OPTION
> 
> DESCRIPTION         [top](https://www.man7.org/linux/man-pages/man1/echo.1.html#top_of_page)
> 
>        Echo the STRING(s) to standard output.
> 
>        -n     do not output the trailing newline
https://www.man7.org/linux/man-pages/man1/echo.1.html

https://www.man7.org/linux/man-pages/man1/touch.1.html

PierreF: touch should also work, since we start from a empty dir (because `touch some_file.txt` unlike `> some_file.txt` don't truncate the file). But in our case it work the same and avoid the odd looking `> some_file.txt`

PierreF (approved):

Issue #935, opened 2026-07-20
Feature req/ suggestion.  Public API for client._out_messages.clear()


JamesParrott:
Minimal example:
```
    def clear_messages(self) -> None:
        with client._out_message_mutex:
            client._out_messages.clear()
```

Power users can encounter situations in which delivery of customised QoS 1 or QoS 2 messages, e.g. using topic aliases fails.  Without any way to clear undelivered messages and try again, these can cause the server to kick the client.  If the user has called client.loop_forever, this creates an unnecessary connection cycle, that they must currently resort to calling Paho's private methods to resolve.

This is the simplest solution suggested by Josh last year in issue #893

Issue #893, opened 2025-06-11
Infinite disconnect/reconnect loops because reconnection resends qos>0 messages even when they're using topic aliasing


inbilla:
# Bug Description

Client will disconnect/reconnect forever in a loop, if the server restarts and a QOS=1 message using topic aliasing wasn't ACKed yet.

The issue appears to occur because upon reconnection, the client will resend all qos > 0 messages that were in flight at the time of disconnection.
However - if one of those messages was using topic aliasing - and set the topic to `""`, then when it is sent to the server (which has restarted and doesn't know the session of the client) the server will close the connection for reason "Protocol Error" (mosquitto). This is because the message had an alias index, but no topic (and the server's state won't know what to do with the message).
The server will now disconnect the client.
The client will reconnect, and the same issue occurs.

NOTE: Probably related to this: https://github.com/eclipse-paho/paho.mqtt.python/issues/855
But I don't think this scenario is covered by that discussion

# Reproduction

Obviously timing this perfectly with a server restart isn't super easy. But recreating the state it ends up in IS.
So here is a quick example script that demonstrates the issue with the current form of the logic.
I reproduced the behaviour using test.mosquitto.org.

```python
import paho.mqtt.client as mqtt


def main():

    def on_connect(client, userdata, flags, reason_code, properties):
        print(f"Connected with result code {reason_code}")

    def on_disconnect(client, userdata, flags, reason_code, properties):
        print(f"Disconnected with result code {reason_code}")

    def on_message(client, userdata, msg):
        print(msg.topic + " " + str(msg.payload), msg.properties)

     def on_log(client, userdata, level, buf):
        print("PAHO-MQTT-LOG:", level, buf)

    mqttc = mqtt.Client(mqtt.CallbackAPIVersion.VERSION2, protocol=mqtt.MQTTv5)
    mqttc.on_connect = on_connect
    mqttc.on_message = on_message
    mqttc.on_disconnect = on_disconnect
    mqttc.on_log = on_log

    # Publish a message using an alias - that was previously registered in an earlier session.
    # If the server has forgotten the client (i.e. a reboot) and we've just reconnected then this message should
    # cause an error on the server because it doesn't know what topic the topicAlias corresponds to.
    alias_property = mqtt.Properties(mqtt.PacketTypes.PUBLISH)
    alias_property.TopicAlias = 1
    mqttc.publish(topic="", payload="hi_aliased", qos=1, properties=alias_property)

[...]

PR #938, opened 2026-07-22
Add clear out messages to public api

JamesParrott:
Workaround for #893 
Includes PR #937 and builds on its new tests. 
Completes #935

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -9,7 +9,7 @@ jobs:
       fail-fast: false
       max-parallel: 4
       matrix:
-        python: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
+        python: ["3.11", "3.12", "3.13", "3.14" ] #, "3.9", "3.10", ]
     runs-on: ubuntu-latest
     timeout-minutes: 5
     steps:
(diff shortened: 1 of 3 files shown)


pswsm (approved):

.github/workflows/tox.yml
```diff
@@ -9,7 +9,7 @@ jobs:
       fail-fast: false
       max-parallel: 4
       matrix:
-        python: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
+        python: ["3.11", "3.12", "3.13", "3.14" ] #, "3.9", "3.10", ]
```
PierreF: Why removing Python 3.9 & 3.10 here ? Tests works on 3.9 on main and on this PR.

JamesParrott: Hi Pierre, 
This was from back in July.  I don't remember exactly why, unfortunately.  Lets try it with 3.9 & 3.10 back in.

JamesParrott: Very possibly I meant to, but simply forgot to put them back in.  

I vaguely remember  Github was only allowing me to run tests in simultaneously on a maximum of 4 python versions at once, and I was getting bored of waiting for the first four to run before testing on 3.13 and 3.14.  

Apologies,

Issue #872, opened 2024-11-27
Client.loop_stop contains a race-condition


PromyLOPh:
# Bug Description

`Client.loop_stop` expects that `self._thread` stays valid, but `_thread_main` unsets `self._thread` after exiting, causing a race-condition:

```bash
$ python3 test.py
Traceback (most recent call last):
  File "test.py", line 14, in <module>
    client.loop_stop()
  File "src/paho/mqtt/client.py", line 2365, in loop_stop
    self._thread.join()
    ^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'join'
```

# Reproduction

Trigger it by applying this patch to paho-mqtt, which simply adds a wait, so `_thread_main` always wins the race:

```patch
diff --git a/src/paho/mqtt/client.py b/src/paho/mqtt/client.py
index 4ccc869..dc01f5e 100644
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -2360,6 +2360,7 @@ class Client:
             return MQTTErrorCode.MQTT_ERR_INVAL

         self._thread_terminate = True
+        time.sleep (2)
         if threading.current_thread() != self._thread:
             self._thread.join()
```

Then run this minimal MQTT client:

```python
import time

from paho.mqtt.client import Client
from paho.mqtt.enums import CallbackAPIVersion, MQTTProtocolVersion

client = Client(CallbackAPIVersion.VERSION2, 'testclient', protocol=MQTTProtocolVersion.MQTTv5)

client.loop_start()
client.connect(host='localhost')

time.sleep (2)

client.disconnect()
client.loop_stop()
```

# Environment

* Python version: Python 3.11.2
* Library version: Commit d45de3737879cfe7a6acc361631fa5cb1ef584bb
* Operating system (including version): Debian 12
* MQTT server (name, version, configuration, hosting details): mosquitto 2.0.15 without any configuration.

JamesParrott:
Hi there, thanks for this.
I wonder if the `self._thread_terminate` checks and `._thread.join()` call are both needed at all?  Won't either one, or the other suffice?  

The `self._thread_terminate` checks might be used by something else starting a thread, that doesn't use `loop_stop` 

I'm not sure what "ghost mentioned this" means, but I've made a draft PR:  https://github.com/eclipse-paho/paho.mqtt.python/pull/940

PR #940, opened 2026-07-26
Remove self._thread.join() (fix race condition in loop_stop)

JamesParrott:
(Over?)simple fix for #872.  

I can't pin it down, but removing the join and relying on a shared boolean feels much less robust.  

What else breaks with this change?

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -2390,8 +2390,6 @@ class Client:
             return MQTTErrorCode.MQTT_ERR_INVAL
 
         self._thread_terminate = True
-        if threading.current_thread() != self._thread:
-            self._thread.join()
 
         return MQTTErrorCode.MQTT_ERR_SUCCESS
 
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -2391,6 +2391,12 @@ class Client:
 
         self._thread_terminate = True
 
+        with self.lock:
+            thread = self._thread
+
+        if thread is not None:
+            thread.join()
+
         return MQTTErrorCode.MQTT_ERR_SUCCESS
 
     @property


PierreF:
Removing the thread.join() means that now loop_stop() no longer wait for thread to terminate before returning which it a good behavior IMO.

This is mostly a TOCTOU issue: we check `self._thread` (`self._thread is None` AND `threading.current_thread() != self._thread`) and later use it (`self._thread.join()`).
We need to perform those action atomically, but since we very likely don't want to hold a lock while call `thead.join()` (holding a lock during long operation is usually a bad idea), what we should probably do is:
```
# Atomically get the value:
thread = self._thread

# Do everything else with thread, not self._thread.
if thread is None:
[...]
```

AFAIK, on CPython, just doing `thead = self._thread` is an atomic variable load (you get a valid value of `self._thread`, no load tearing possible). But I believe there is no guaranteed on other implementation of Python. So using a lock is probably preferable:
```
with some_lock():  # and all read & write access to self._thread need the same locks
  thread = self._thread
```

If implementing with the lock cause issue, we might rely on CPython having atomic variable load

JamesParrott:
Even if the atomic assignment is reliable, moving towards locks is probably the way forward, as free threaded builds of Python become more mainstream (and even GIL free variants).

However, as I understand it, the thread does get closed regardless - the error in #872 is just a nuisance for the user, from a method call on None.  To avoid that, instead of deleting the call altogether, we could simply do:

```py
if threading.current_thread() != self._thread and self._thread is not None:
            self._thread.join()
```

PierreF:
AFAIK the last version is incorrect in any Python version. The thread could switch between any byte code instruction. So it's highly likely that:
```
if threading.current_thread() != self._thread and self._thread is not None:  # <-- produce multiple instructions; but don't matter that much
    # --- Here thead could switch: we finished instructions from the if line. The issue is THIS thread switch
    self._thread.join()
```

The maybe valid code, with CPython because AFAIK thread switch only occur between bytecode instruction (GIL or not; probably free threaded or not)
```
local_variable = self._thread
if local_variable is not None:
   pass
```

Because the above produce the following bytecode:
```
  2           LOAD_FAST_BORROW         0 (self)
              LOAD_ATTR                0 (_thread)   # <<<< Therefore this LOAD_ATTR is atomic. We atomically read _thread
              STORE_FAST               1 (local_variable)

  3           LOAD_FAST_BORROW         1 (local_variable)
              POP_JUMP_IF_NONE         3 (to L1)
              NOT_TAKEN

  4           LOAD_CONST               0 (None)
              RETURN_VALUE

  3   L1:     LOAD_CONST               0 (None)
              RETURN_VALUE
```

But once more, because this rely on:
- Python having (this) bytecode (true on CPython, wrong on PyPy, Jython, IronPython - even if some might be dead)
- That Cpython only switch thread between bytecode instruction - not while executing one. Which I'm not sure Python ever committed to kept this forever

That why I think the best and correct version is:
```
with self.lock:
  thread = self._thread

if thread is not None:
  thread.join()
```

With this (and because all other read/write access to `self._thread` must also be done with the lock), it's guarantee that the value on which we call `.join()` is a `Thread` (because it's not None).
So the only point of this solution is: we must be able to change all access to `self._thread` to be under the same lock.

JamesParrott:
Thanks Pierre - I fully admit I've only considered this from a CPython-centric view point.  And I hadn't appreciated that the reference to the thread itself could change.  I'll include your suggestion, instead of just dropping the join call.

JamesParrott:
Oops!  Sorry, I was a bit hasty there, and took the code suggestion literally (there is no `self.lock`, and no `self._lock` either).  To make acquiring a lock work, we need to decide which of the 7 we currently create, to acquire:

```
        self._in_callback_mutex = threading.Lock()
        self._callback_mutex = threading.RLock()
        self._msgtime_mutex = threading.Lock()
        self._out_message_mutex = threading.RLock()
        self._in_message_mutex = threading.Lock()
        self._reconnect_delay_mutex = threading.Lock()
        self._mid_generate_mutex = threading.Lock()
 ```

PierreF:
None of current lock name seems to be a good fit. The closest is _reconnect_delay_mutex, but then we should drop delay.
I would say, since we already have close to one lock per action, create a new lock dedicated to this. Unless we need to take two locks, have small dedicated lock seems a better approach.

PR #941, opened 2026-08-04
fix: honor Server Keep Alive from MQTT 5.0 CONNACK

### LINKED ISSUE #905: keepalive cannot be set after connect
Keepalive cannot be set after connect (e.g. in the on_connect callback). In MQTTv5 the server can answer the connect with a property of the maximum keep alive that it accepts. I would want to lower the keep alive in on_connect to this value if necessary, but it is not possible anymore after the socket is open.

Also, wouldn't it make sense for the base client to deal with all the v5 properties (keepalive, max_packet_size, etc.) the server can send and set the client values accordingly?

Puneet15556:
Fixes #905

## Problem

Per MQTT 5.0 §3.2.2.3.14, when the server returns a **Server Keep Alive** in the CONNACK, the client MUST use that value instead of the keepalive it sent. paho did not apply it, so ping timing kept using the client's original value and users had to work around it by setting keepalive in `on_connect` (which the setter blocks anyway).

## Change

In `_handle_connack()`, after unpacking the CONNACK properties, apply the Server Keep Alive when present:

```python
if hasattr(properties, "ServerKeepAlive"):
    self._keepalive = properties.ServerKeepAlive
```

This handles it automatically inside paho, as suggested by @JamesParrott in #905. `self._keepalive` is what the ping scheduler uses, so subsequent PINGREQ timing follows the server's value.

## Notes

- Only affects the MQTT v5 path; v3.x is unchanged (no properties).
- `hasattr` guards the case where the server omits the property.
- Happy to add a test using the v5 test harness if you'd like one.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -3891,6 +3891,11 @@ class Client:
                 reason = ReasonCode(CONNACK >> 4, identifier=result)
                 properties = Properties(CONNACK >> 4)
                 properties.unpack(self._in_packet['packet'][2:])
+                if hasattr(properties, "ServerKeepAlive"):
+                    # MQTT 5.0: when the server returns a Server Keep Alive in
+                    # CONNACK, the client must use that value instead of the one
+                    # it sent (MQTT 5.0 spec 3.2.2.3.14).
+                    self._keepalive = properties.ServerKeepAlive
         else:
             (flags, result) = struct.unpack("!BB", self._in_packet['packet'])
             reason = convert_connack_rc_to_reason_code(result)
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -9,6 +9,7 @@ from paho.mqtt.packettypes import PacketTypes
 from paho.mqtt.properties import Properties
 from paho.mqtt.reasoncodes import ReasonCode
 
+import tests.mqtt5_props as mqtt5_props
 import tests.paho_test as paho_test
 
 # Import test fixture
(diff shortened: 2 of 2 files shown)


JamesParrott:
LGTM.  Great stuff - well done.  

Note. The lint_python workflow failing is not due to this PR (so that's not your fault Puneet - don't worry about that).  I'm guessing a third party dependency in the build pipeline (that perhaps we can better control?), probably in the Github Actions runners again, has changed.  That workflow is erroring due to issue #942 (probably nltk v3.10.1)

Re: tests.  Adding a test is always a good idea.  But in this PR, the change are simple, and pretty much entirely consists of the exact same change users must manually carry out themselves, so personally I don't think this fix should be delayed until a broker test is added.  

Hopefully it's straightforward to add a one off test for the client though - set one up with some keep alive value other than 60, and assert it is 60 after connection (changing server keep alive in a broker test would also rule out something odd for that specific value - the client's default keep alive value is also 60).  It's up to you.

If you (or if anyone) can figure out a good way to configure Server keep alive in the broker-in-the-loop tests, that would be superb.  Perhaps all that is required are some `**kwargs` adding to `start` and `run` (and passing them), and then in `run`doing `options.update(kwargs).  But [it doesn't seem very accessible so far](https://github.com/eclipse-paho/paho.mqtt.testing/blob/9d7bb80bb8b9d9cfc0b52f8cb4c1916401281103/interoperability/mqtt/brokers/start.py#L155) without a separate PR to that test repo.

Puneet15556:
Thanks @JamesParrott! I've added a client-level test (`Test_connect_v5::test_connack_server_keep_alive`) rather than touching the broker-in-the-loop harness, since as you noted that isn't easily configurable yet.

It connects a v5 client with the default keepalive of 60, has the fake broker return a CONNACK carrying a Server Keep Alive of 42 (a value other than the default so the override is unambiguous), and asserts `_keepalive == 42` after connection. I confirmed it fails without the fix (`assert 60 == 42`) and passes with it.

JamesParrott:
Excellent.  Really good work @Puneet15556

JamesParrott (approved):
Ship it!

PierreF:
Thanks for your contribution & thanks James for the review.

Feedback from my memory: I believe we moved / added new test using the fake_broker (as you added in your test), mostly because the other option (broker-in-the-loop) wasn't flexible enough / too complex to build specialized test for specific use case. So I believe that your added test are at the best place.

Issue #873, opened 2024-11-28
open connection with custom socket


LibraryRaven:
# Feature Description

I would like to be able to connect with a manually provided socket.

This would enable me to tunnel MQTT through an SSH connection with [paramiko](https://docs.paramiko.org/en/stable/api/transport.html#paramiko.transport.Transport.open_channel).

```python3
import paramiko

# connect to SSH gateway
sshclient = paramiko.SSHClient()
sshclient.connect("sshgateway.example.org", "user", "password")

def create_socket():
  # open tunnel to the internal MQTT port
  return sshclient.get_transport().open_channel('direct-tcpip', ("127.0.0.1", 1883), ('', 0))

import paho.mqtt.client as mqtt
mqttc = mqtt.Client(mqtt.CallbackAPIVersion.VERSION2)
mqttc.connect("127.0.0.1", transport="callback", create_socket=create_socket) 
```

# Requested Solution

```python3
import socket

def create_socket():
  sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
  sock.connect(('mqtt.example.org', 1883))
  return sock

import paho.mqtt.client as mqtt
mqttc = mqtt.Client(mqtt.CallbackAPIVersion.VERSION2)
mqttc.connect("mqtt.example.org", transport="callback", create_socket=create_socket) 
```

# Alternatives

For now, I can use my own SocketedClient class, which inherits Client and uses a provided create_socket function on connect() and reconnect().

JamesParrott:
Hi @LibraryRaven,
I've added a draft PR. It pretty much just formalises support in the API for what you're doing already, but dynamically changing _create_socket on each instance (instead of subclassing Client, or adding a new arg to connect).  I can't promise it'll get merged, but lets see what the others think

In the meantime, I've not tested this myself, but Paho.MQTT.Python [already supports SOCKS Proxies](https://eclipse.dev/paho/files/paho.mqtt.python/html/client.html#paho.mqtt.client.Client.proxy_set), and AsyncSSH supports [dynamic port forwarding via SOCKS](https://asyncssh.readthedocs.io/en/latest/api.html#asyncssh.SSHClientConnection.forward_socks). Have you considered using AsyncSSH instead of Paramiko (the latter can be a bit fiddly anyway)?

If that doesn't work, it feels to me like the kind of thing Paho should be able to do right now, and would be more of a bug fix than a feature request.

PR #945, opened 2026-08-08
Add property & setter: Client.create_socket

JamesParrott:
Implements #873 

I'm raising this as a draft PR now to gather feedback.  I'll add tests in due course.

### DIFF
--- a/src/paho/mqtt/client.py
+++ b/src/paho/mqtt/client.py
@@ -853,6 +853,7 @@ class Client:
         self._port = 1883
         self._bind_address = ""
         self._bind_port = 0
+        self._custom_create_socket: Callable[[], SocketLike] | None = None
         self._proxy: Any = {}
         self._in_callback_mutex = threading.Lock()
         self._callback_mutex = threading.RLock()
@@ -1102,6 +1103,16 @@ class Client:
     @logger.setter
     def logger(self, value: logging.Logger | None) -> None:
         self._logger = value
+
+    @property
+    def create_socket(self) -> Callable[[], SocketLike]:
+        if self._custom_create_socket is None:
+            return self._create_socket
+        return self._custom_create_socket
+
+    @create_socket.setter
+    def create_socket(self, val: Callable[[], SocketLike]) -> None:
+        self._custom_create_socket = val
 
     def _sock_recv(self, bufsize: int) -> bytes:
         if self._sock is None:
@@ -1608,7 +1619,7 @@ class Client:
                 if not self.suppress_exceptions:
                     raise
 
-        self._sock = self._create_socket()
+        self._sock = self.create_socket()
 
         self._sock.setblocking(False)  # type: ignore[attr-defined]
         self._registered_write = False


PierreF:
The issue I see with adding API to simply replacing `_create_socket` is that this function does lots of task:
* It connect to unix OR TCP server
* It select the host & port to connect to (and the timeout and the source address). If multi-addresses broker support is added it might cause conflict here
* It handle proxy (HTTP & SOCKS)
* It handle TLS/SSL
* It handle websocket

So we need to carefully pick the API we use. I believe we shouldn't replace the full `_create_socket` as it would either means:
- we can't assume all setup listed above are done, which might cause other issue (lots of attribute, `self._host`, `self._transport`, ... can't be trusted to be right)
- user provided version must implement all of this

But maybe the right level would be to be a "proxy" level:
- Either user set a "classic" proxy (or no proxy) - today behavior
- Or use a function proxy, which create the TCP connection. So it's the function that replace `socket.create_connection` with all its arguments

This way most feature of `_create_socket` are kept, and I think it make sense to be at "proxy" level for this feature.

I think think to a `set_create_connection()` which take a function `create_connection(addr, timeout=self._connect_timeout, source_address=source)`. It's documented to only be used for for TCP connection (not unix socket) and that TLS/SSL and websocket is added after (which is not configured isn't done).

Does this sound good to you ?

JamesParrott:
Thanks for your thoughts, Pierre. 

I fully understand where you're coming from - _create_socket can call to three sub methods (and _WebsocketWrapper), giving 6 possible combinations (unix, tcp & websockets, all three both with and without ssl).  Plus possibly handling a proxy.

But, as far as I can tell (please correct me if I'm wrong), that method is essentially a factory that returns one of 10 possible combinations.  None of those methods alter the client's state.  Given a few extra args for the client's attrs, they could all be refactored into functions that return some socket (I'm not suggesting this).  `_proxy_is_valid` is already a static method.

My idea is, that instead of in future adding even more config and custom logic, to implement an 11th, a 12th, or even a 20th socket, lets just let those users that want those sockets, create them and supply them themselves.  All they need to do it implement the SocketLike.

I completely agree that any "user provided version must implement all of this".  The intended user base for this feature would be required to handle all that for themselves, whilst creating their custom socket.  And LibraryRaven for one, is telling us that they're not only willing to do so, they have already done this.  And perhaps other third party libraries, as well as Paramiko even provide socket instances out of the box already, that with a few tweaks, Paho.mqtt.Python should happily work with.

I'll bow to your experience though, if you think officially supporting this would create even more issues and headaches from such power users in future.

PierreF:
The issue I see with replacing `_create_socket` with a user-provided function, is that either:
- we have multiple attribute that are likely wrong (`self._host`, `self._ssl`, `self._transport`, `self._tls_insecure`).
- or the user-provided function will need some hack (access to private field or pass them twice hoping nothing de-synchronize them)

My initial idea is that the provided user-function should be able to replace a existing function that don't access private fields, hence my suggestion to connect to "proxy-level".

On the other hand, I agree that being at "SocketLike" interface is better then being a "TCP proxy opened" as it avoid to assume TCP. It opens to more possibility.

But I think a good interface should allow to re-implement the internal paho socket creation with a user-provided implementation. For example to reimplement `_ssl_wrap_socket` but using another SSL implementation. Which means that `_ssl_wrap_socket` should be implementable without accessing private field/method. Possibly passing an argument to the `create_socket` function which contains all details ?

My idea could be a signature like `create_socket(info: ConnectionInfo) -> SocketLike` with `ConnectionInfo` being a structure with the host, port, ssl + ssl details (insecure, ssl_context), proxy hints ?. This way, the user-implementation could works exactly as paho's implementation, likely because the updated paho implementation could access data from the `info` parameter rather than from a `self._variable`.

Going this way, I would probably include two other point:
- the `_ssl_wrap_socket` might be useful to user-implementation. In other to prove our interface, I would move it to plain function `def ssl_wrap_socket(bottom_sock: SocketLike, info: ConnectionInfo) -> SocketLike`. This way paho would use it but user-implementation could also.
- the `self._transport` should be `"custom"` (or the function itself ?) when a custom connection factory is provided. This would solve the inconsistency of `self._transport` that could be `"websockets"` when you actually have an tcp+ssh. I'm fine with changing `self._transport` to `"custom"` when setting `self.create_socket` or that changing the create_socket goes through the `self.transport` setter (the later is a bit odd in that the setter would take a `str | SocketFactory`).

Obviously with this interface, the user create_socket have access to all details (host, port, ssl settings) and *could* use them. If they don't, it's up to the user. But they have everything in their hand to handle / validate them (e.g. apply the settings, silently ignore them even if it might cause unexpected result, or maybe raise error for unsupported settings).

PR #955, opened 2026-09-16
Fix MQTT 5 disconnect reason parsing for short packets

### LINKED ISSUE #860: Incorrect handling of DISCONNECT packet
I believe the handling of DISCONNECT (at least for MQTT 5) isn't correct:

https://github.com/eclipse/paho.mqtt.python/blob/d45de3737879cfe7a6acc361631fa5cb1ef584bb/src/paho/mqtt/client.py#L4021C1-L4022C1

I'm using paho as a test client and it doesn't seem to be correctly handling a DISCONNECT packet with a non-zero reason code and zero length properties (i.e. Remaining_Length = 2).

Shubham-Padkonde:
Fixes #860.

MQTT 5 DISCONNECT packets with a one- or two-byte Remaining Length lose their reason code. For example, `e0 02 87 00` reports a normal disconnection through callback API v2 instead of "Not authorized".

Read the reason code when Remaining Length is at least one, and the property length when it is at least two, matching [MQTT 5 sections 3.14.2.1 and 3.14.2.2](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html). Keep the existing callback behavior when fields are omitted.

Tests send real packets through the existing fake broker and inspect both callback APIs. They cover omitted fields, an explicit normal reason, a nonzero reason without properties, an empty property section, and a Reason String. Both TCP and Unix sockets run on Linux.

Validation:
- Regression on Windows/Python 3.12: 5 failed and 5 passed before the fix; all 10 passed after.
- Full `tox -e py311` on Ubuntu 22.04/Python 3.11.15: **188 passed, 2 xfailed**. Used the testing repository revision pinned by CI.
- `tox -e lint`: passed, including all required pre-commit hooks and mypy. The configured nonblocking Black/codespell checks still report existing repository-wide findings.
- Changed-file Ruff and `git diff --check`: passed.

The Linux suite was run from a native Linux filesystem: the initial run under the Windows-mounted directory could not create Unix-domain sockets. No project changes were made for that environment limitation.

Prepared with OpenAI Codex (GPT-6) assistance, including implementation, tests, and this description. Eclipse contributor agreement status still needs verification before acceptance.


JamesParrott (changes requested):
This PR has not been tested in anywhere that the results are public, as far as  I can see.

Shubham, please "Enable Workflows" on your fork, and make a trivial commit on top of this PR to trigger the test job (and others).

Shubham-Padkonde:
Thanks for the review. I enabled Actions on the fork and pushed an empty, signed-off commit to trigger the existing workflows without changing the patch. Public runs are now available:

- Python 3.9-3.14 test matrix: https://github.com/Shubham-Padkonde/paho.mqtt.python/actions/runs/35151860773
- Lint: https://github.com/Shubham-Padkonde/paho.mqtt.python/actions/runs/35151860715
- Pre-commit: https://github.com/Shubham-Padkonde/paho.mqtt.python/actions/runs/35151860635

They are still running; I will check the results.

Shubham-Padkonde:
The public fork run is now green across Python 3.9, 3.10, 3.11, 3.12, 3.13, and 3.14: https://github.com/Shubham-Padkonde/paho.mqtt.python/actions/runs/35151860773 . The lint and pre-commit runs linked above also passed.

JamesParrott:
That's really helpful - thankyou :).

JamesParrott:
In the spirit of TDD (albeit in the wrong order), I copied and pasted the test method from this PR to a clean branch.  And it does [fail](https://github.com/JamesParrott/paho.mqtt.python/actions/runs/35199484852/job/105130534561) as expected, so the test is definitely non-trivial.  

The relevant part of the v5 spec is [here](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901205).  Also:
  in the [corresponding part of the v3.1.1 spec](https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/errata01/os/mqtt-v3.1.1-errata01-os-complete.html#_Toc442180903) servers do not send disconnect packets (only clients do) so `_handle_disconnect` shouldn't be called if only using MQTT 3.
The original thresholds that this PR changes, were first set in [this commit](https://github.com/eclipse-paho/paho.mqtt.python/commit/2e6accccfc5db74ea9f918ad990ca6a3ead0087e)

JamesParrott:
OK, so overall, I'll trust this does fix issue 860, and I've personally not been able to find anything wrong with this PR.  But I don't understand the packet unpacking logic deeply to say for sure it won't break anything else (I couldn't find a new packet header handling method that made the old thresholds obsolete).  Nor to say if the Fake Broker (that this is tested against) is sending the correct packets at all.  

In particular, similar threshold tests to the one this PR changes against `self._in_packet['remaining_length']` occur in quite [a](https://github.com/eclipse-paho/paho.mqtt.python/blob/26cf3e3503bfde014f7c67dd12f4bed23f74a575/src/paho/mqtt/client.py#L4311) [number](https://github.com/eclipse-paho/paho.mqtt.python/blob/26cf3e3503bfde014f7c67dd12f4bed23f74a575/src/paho/mqtt/client.py#L4255) of other places.  If this PR is merged, shouldn't we take another look at those too?  There are various other issues for PUBACK and PUBREL etc.  

It's probably  naive to assume all the other values are off by 2 too, but it would be good to understand why they aren't.

But despite that, this PR looks like good work.  Thanks @Shubham-Padkonde

Shubham-Padkonde:
Thanks for independently checking the regression. The two-byte difference comes from the packet identifier, rather than a change to the shared packet reader.

_packet_read decodes Remaining Length separately and stores only the following bytes in _in_packet['packet']. In _handle_pubrel, _handle_pubrec and _handle_pubackcomp, the first two bytes are unpacked as the packet identifier. Their optional reason code therefore starts at offset 2 (remaining_length > 2), and properties at offset 3 (> 3).

DISCONNECT has no packet identifier. Its reason code starts at offset 0 and its property length at offset 1, which is why this PR uses > 0 and > 1. The fake broker sends the complete wire packet: E0, the encoded remaining length, then the test payload. For these short payloads the length fits in one byte. The test goes through the socket reader, rather than calling the handler directly.

I checked these handler offsets in the current source; this is not a claim that every other MQTT packet path is free of bugs. I would keep separate PUBACK/PUBREL issues scoped to their own reproductions.

Prepared with Codex assistance.

PierreF:
> Eclipse contributor agreement status still needs verification before acceptance.

As you initially said, the [ECA](http://www.eclipse.org/contribute/cla) need to be signed for your contribution to be accepted (the [wiki](https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#Eclipse_Contributor_Agreement) explain steps to sign it).

Issue #957, opened 2026-09-21
Update CONTRIBUTING to reflect AI policy


pswsm:
- [x] update the Issue templates to include a section for the AI generated content.
- [x] update `CONTRIBUTING.md` to add Eclipse's AI guidelines

Feel free to add more things related. This what came up in the mailing list.

MattBrittan:
I asked Codex if it would read `CONTRIBUTING.md`, it said no but suggested creating an `agents.md` file with the following (it states that Claude code will also read this) would resolve this. Note that this was run against the Go repo and I can't see a pull request template here, it might be worth creating one (the Go one is [here](https://github.com/eclipse-paho/paho.golang/blob/master/PULL_REQUEST_TEMPLATE.md) - it's far from perfect but I hope it reiterates the need to sign the ECA).

Update: I've added an [AGENTS.md](https://github.com/eclipse-paho/paho.golang/blob/master/AGENTS.md) to the Go v5 client.

# Contribution workflow

Before planning or implementing a contribution, read and follow
CONTRIBUTING.md and PULL_REQUEST_TEMPLATE.md.

Follow their requirements for finding an existing issue, discussing
larger changes, contributor eligibility, and testing.

When preparing a pull request, use PULL_REQUEST_TEMPLATE.md,
reference the related issue, and describe the testing performed.

pswsm:
Well thanks, I'll check it out.

PierreF:
I think the PULL_REQUEST_TEMPLATE.md could also be added, it allows to remind key point (like the ECA). On this point, shouldn't the PULL_REQUEST_TEMPLATE.md be in .github folder for Github to use it ? (paho.golang have it at root, does Github use it there @MattBrittan ?)

pswsm:
> I think the PULL_REQUEST_TEMPLATE.md

For sure, I'm still planning to add the AGENTS.md file.

MattBrittan:
> paho.golang have it at root, does Github use it there @MattBrittan ?

It does use it currently, but I think `.github` is a better location ([docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates)). The Go v3 client has it under `.github`, not sure why the V5 client did things differently (predates my involvement). Edit: The V5 client followed the [C client structure](https://github.com/eclipse-paho/paho.mqtt.c/blob/master/PULL_REQUEST_TEMPLATE.md).

[...]

PR #958, opened 2026-09-21
docs: contrib: update

pswsm:
I've copied the golang's CONTRIBUTING.md file and changed the Go reference's to Python.

If it's too much of a change, I'll change it back.


pswsm:
Forgot to mention issue #957

CONTRIBUTING.md
```diff
@@ -1,116 +1,57 @@
...
 Create a new bug:
 -----------------
 
-Be sure to search for existing bugs before you create another one. Remember
-that contributions are always welcome!
-
-- [Create new Paho bug](https://github.com/eclipse/paho.mqtt.python/issues)
-
-
-## Committer resources:
```
PierreF, line 87: We lost this information. But I agree it don't belong to CONTRIBUTING.md.

Ideally most should be moved to Github action (but not all, e.g. the GPG signing done by developer is still a manual action that need to be documented). Fine for now, we could always recover this information from git history.

PierreF (approved):

CONTRIBUTING.md
pswsm: Okay, I can add it back no problem. I admit I was a bit brute by overwriting the whole file.

JamesParrott (commented):
LGTM

